PR #17, opened 2015-04-02
Fix the beginning-of-error cnum off-by-2

dsheets:
The column position of the start column always seemed to be 2 more than the actual 0-indexed column. I'm not entirely certain of the reasoning behind the offsets but this appears to fix the problem for me. :-/


mjambon:
Ok, I believe you. I usually don't rely on positions in json data because it's impractical (instead I look at the sample data included in the error message), so that must be why it has never been caught before.

PR #19, opened 2015-04-27
Added a keys helper function to util module

fxfactorial:
This PR adds a keys util function; when playing around with json, its nice to see a high level view.


mjambon:
Great. Thanks.

PR #21, opened 2015-09-24
Two new utilities, remove unneeded function

fxfactorial:


util.mli
```diff
@@ -68,13 +68,15 @@ exception Undefined of string * json
       return undefined. Currently this only happens when an array index is out
       of bounds. *)
 
-val ( |> ) : 'a -> ('a -> 'b) -> 'b
-  (** Forward pipe operator; useful for composing JSON access functions
-      without too many parentheses *)
```
mjambon: Removing it may break compatibility for some people, e.g. `let (|>) = Yojson.Util.(|>)`. We can instead mark it as obsolete, and define the value as `let (|>) = (|>)` in the ml file. Otherwise the rest looks good to me. Thanks!

fxfactorial: @mjambon Ah, I didn't think of that. I'm not sure how you meant by marking it obsolete, like by putting it in the doc string or I've seen some people do the ppx thing, something like `[@@ocaml.deprecated "blah blah"]`.

util.mli
```diff
@@ -72,9 +72,15 @@ val ( |> ) : 'a -> ('a -> 'b) -> 'b
   (** Forward pipe operator; useful for composing JSON access functions
       without too many parentheses *)
```
mjambon: I haven't tested this but ocamldoc supports `@deprecated`. I guess you'd place it at the beginning of the comment or something. See http://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html

PR #23, opened 2016-01-22
Add a Yojson.Safe.Util module

emillon:
Hi,

Is there a reason for the `Util` submodule to be defined only for the `Basic` variant?

In case there is not, here is a patch that adds a `Util` submodule to the `Safe` variant, similar to the one defined in `Basic`.

Thanks!

### DIFF
--- a/util.ml
+++ b/util.ml
@@ -8,6 +8,9 @@ let typeof = function
   | `List _ -> "array"
   | `Null -> "null"
   | `String _ -> "string"
+  | `Intlit _ -> "intlit"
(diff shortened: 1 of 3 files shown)


mjambon:
I don't remember a good reason for not doing this.

Sorry for the delay!

mjambon:
and thank you @emillon !

mjambon:
Should be available from opam soon as yojson 1.3.1 https://github.com/ocaml/opam-repository/pull/5677

PR #25, opened 2016-02-21
Cleanup and addition of a formatter function

### LINKED ISSUE #24: use standard types for IO
Would it be possible to provide `Format` compatible printers directly, such as `Yojson.Safe.print : Format.formatter -> json -> unit`? It is more convenient and direct than using easy-format, and would possibly allow to get rid of the dependency on easy-format…

c-cube:
Should resolve #24. I also removed trailing spaces and tabs (can revert if you want).
Are there plans for making yojson and its dependencies `-safe-string` compatible?

### DIFF
--- a/common.ml
+++ b/common.ml
@@ -87,7 +87,7 @@ type lexer_state = {
     (* Current line number (starting from 1) *)
 
   mutable bol : int;
-    (* Absolute position of the first character of the current line 
+    (* Absolute position of the first character of the current line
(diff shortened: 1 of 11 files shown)


mjambon:
Thanks!

mjambon:
Should be in opam soon as yojson 1.3.1 https://github.com/ocaml/opam-repository/pull/5677

PR #35, opened 2016-10-26
Fix lexing failure

314eter:
When there is an invalid token consisting of one character followed by eof, `long_error` tries to print more than one character, which results in a failure.

```ocaml
# Yojson.Safe.from_string "x";;
Exception: Failure "lexing: empty token".

# Yojson.Safe.from_string "xx";;
Exception: Yojson.Json_error "Line 1, bytes 0-2:\nInvalid token 'xx'".
```

I don't know whether there was a valid reason to look for more characters. If not, the fix is very easy.


mjambon:
We're only using the `junk` pattern and rule for error reporting so this should work.

mjambon:
opam release in progress https://github.com/ocaml/opam-repository/pull/7691

PR #39, opened 2017-03-19
Fix link

karlhorky:


mjambon:
Thanks.
That page won't be updated, so we'll have to find a way (i.e. volunteers) to produce and host the ocamldoc documentation.

karlhorky:
No problem! Thanks for this project.

Hm, what would be involved in this work? Could it be automated somehow?

mjambon:
It would be good if opam took care of building and hosting the html documentation for all the packages from the official [opam-repository](ocaml/opam-repository). It would be a much more rewarding effort than managing documentation hosting for each project separately.

I'm not sure what's blocking at the moment (lack of volunteers? excessive requirements?). I haven't thought deeply about the question, but here's the list of features I can think of:

Very important and valuable:
* documentation automatically built and hosted for the opam package maintainers
* package maintainers get notified when the documentation can't be built
* supports at least ocamldoc-generated HTML documentation
* hosting for at least the latest buildable documentation of each official opam package (opam-repository)
* an index page with a link to the documentation of each package

Nice-to-have but not essential:
* documentation for each minor version of each package, not just the latest
* support for arbitrary html bundles, not just ocamldoc-generated
* built-in search (for fresher results and support for operators, and other OCaml-specific details, as opposed to a Google "site:" search)
* a cookie-based list of the recently-visited packages, for the developer to quickly find what they're looking for
* simple redirects to external websites, for authors who prefer to host the package documentation themselves or don't want the same documentation to exist in two places

Not important at all:
* uniform styling and formatting of the documentation for each package
* extensibility with custom opam repositories and alternate hosting site

It's more a question for the opam developers. Let's see.

karlhorky:
Wow, that sounds comprehensive! Maybe we can rustle up some interest from interested parties.

mjambon:
I asked on Twitter, to get a sense of what the current situation is https://twitter.com/mjambon/status/843944170087563264

PR #40, opened 2017-05-22
Use cppo.ml suffix for preprocessed files

olafhering:
The documentation of jbuild and oasis/ocamlbuild states that files
ending in .cppo.ml will be automatically processed based on the
extension. Rename source files to allow build with jbuild/_oasis.

Signed-off-by: Olaf Hering <olaf@aepfle.de>


mjambon:
Thanks!

PR #41, opened 2017-07-01
Jbuilder port

rgrinberg:
Now yojson installs .annot files, supports bytecode only, and all the other
goodies that come up with letting jbuilder handle this stuff for you.

Also, I've elected to get rid of yojson.biniou. Packaging it without its own
opam package is painful and unnecessary as it has no dependencies of its own.
Just putting the Yojson_biniou also has no real disadvantages as OCaml will
simply not link this module if it isn't used in user code.


mjambon:
Thanks a lot.

PR #43, opened 2017-07-18
Enforce -safe-string

jpdeplaix:
Sorry, yet another safe-string related PR.. :/
It's just to enforce for future developments and also I realized the ```bytes``` dependency is pretty useless because yojson is only available only for OCaml >= 4.02 (where the ```Bytes``` module exists anyway).

### DIFF
--- a/bench/jbuild
+++ b/bench/jbuild
@@ -3,4 +3,5 @@
 
 (executable
  ((name bench)
+  (flags (-safe-string))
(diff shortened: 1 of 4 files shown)


mjambon:
Great. I'm releasing this.

mjambon:
opam-repository PR https://github.com/ocaml/opam-repository/pull/9868

PR #52, opened 2018-02-18
Remove .merlin/INSTALL/META.in instructions

rgrinberg:
Installation is done through opam as it's done for all OCaml packages and the
rest are generated by dune now.


mjambon:
Thanks.

PR #53, opened 2018-02-19
Add back install/uninstall targets to Makefile

ArchangeGabriel:
Same as https://github.com/mjambon/easy-format/pull/20.

I’ve also removed a duplicated `bench` in `.PHONY`.


mjambon:
Great. Thanks!

PR #55, opened 2018-03-28
Switch to topkg

rgrinberg:
This PR switches the release process to use topkg. This makes releasing much more less manual and more consistent. Some things to keep in mind:

* We no longer need to use cppo to substitute version numbers
* To actually make the github submission, we will need to use opam publish. Due to unfortunate circumstances, opam-publish is only available on OCaml 4.05.0. So if you have a 4.05.0 switch then great. Otherwise, you will have to do the last step manually.

The release process with topkg is as follows:

```
topkg tag # tags a release using the version number in the CHANGELOG.md
topkg distrib --skip-build --skip-lint --skip-tests # creates a tarball that will be the release
topkg publish distrib --verbose # uploads the tarball the github
topkg opam pkg # creates the local opam package submission in _build
topkg opam submit # requires opam publish. if you don't have it then simply copy paste the files in _build into your local opam-repository and make the PR yourself.
```

Note that it's important to create our own tarball here. Github likes to change the archives they host which correspond to git tags subtly, and hence invalidate the checksums. Uploading our own tarballs protects against this tampering.


mjambon:
Thanks for the patch and for the instructions.

I haven't tried topkg yet beyond `topkg tag` (for the atd repo). If you need to make a release urgently, you can proceed, otherwise I'd prefer to do it myself so I learn something.

(my schedule is packed with other things; not sure when I'll get to it)

rgrinberg:
There's no rush. It would be ideal if we both us did the releases the same
way so take your time and feel free to ask.

PR #57, opened 2018-07-10
fix jbuild dependencies for cppo generation

avsm:
Without this fix, there is an occasional build race on trying
to find `type.ml`:

```
Error: File "yojson.cppo.mli", line 43, characters 0-19
Error: Cannot find included file "type.ml"
```

Also remove a duplicate dependency on `type.ml` in the previous
rule as part of this PR.

Signed-off-by: Anil Madhavapeddy <anil@recoil.org>

### DIFF
--- a/lib/jbuild
+++ b/lib/jbuild
@@ -9,7 +9,6 @@
     read.ml
     write.ml
     safe.ml
-    type.ml
     pretty.ml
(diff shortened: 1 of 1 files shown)


mjambon:
Well-spotted, thanks!

Issue #61, opened 2018-09-13
Failure "oops" from Yojson.Safe.to_channel


NathanReb:
I recently hit the `Failure "oops"` raised at https://github.com/mjambon/yojson/blob/master/lib/write.ml#L31.

I unfortunately haven't been able to reproduce it in a minimal example so far but here are the debug log and the `yojson` part of the stacktrace:

```
src="CKA_ALWAYS_SENSITIVE" start=0 len=20
```
```
Called from file "write.ml", line 52, characters 2-24
Called from file "write.ml", line 329, characters 4-21
Called from file "write.ml", line 284, characters 6-15
Called from file "write.ml", line 334, characters 2-24
Called from file "write.ml", line 290, characters 6-15
Called from file "write.ml", line 339, characters 2-29
Called from file "write.ml", line 284, characters 6-15
Called from file "write.ml", line 339, characters 2-29
Called from file "write.ml", line 290, characters 6-15
Called from file "write.ml", line 334, characters 2-24
Called from file "write.ml", line 290, characters 6-15
Called from file "write.ml", line 334, characters 2-24
Called from file "write.ml", line 290, characters 6-15
Called from file "write.ml", line 339, characters 2-29
Called from file "write.ml", line 290, characters 6-15
Called from file "write.ml", line 334, characters 2-24
Called from file "write.ml", line 457, characters 2-21
```

The error seems unrelated to the `Yojson.Safe.json` value which I can write to a file or string without any problem or even to a seemingly identical `out_channel` in utop.

The commit that triggers this exception doesn't change either the `open_out_gen` or the `Yojson.Safe.to_channel` calls so it's a bit of a nightmare to debug.

How about making this error more explicit, for example by reraising the caught exception instead of the `failwith` ?

NathanReb:
Just to clarify: I don't think it's a bug from `yojson` but it could use a better error reporting, happy to contribute with a PR for the reraise if you feel like that's the right way to go!

mjambon:
Oh, that's awful. Sorry about that.

A first step would be to print the exception raised by `Bi_outbuf.add_substring` as well as the state of the output buffer (the `int` fields in https://github.com/mjambon/biniou/blob/master/src/bi_outbuf.mli#L3-L40). The workflow for this is to `git clone` yojson, edit it, and `opam pin` it so as to use this version instead of the one coming from the opam repository. That will take care of rebuilding all the opam packages that depend on it. Then you can rebuild your application and see what's going on.

[...]

PR #64, opened 2018-10-25
Don't ignore exceptions

### LINKED ISSUE #62: Don't ignore exceptions
#61 was a bug that was hard to fix because an exception coming from another library was not reported and instead `Failure "oops"` (sic) was raised.

This task is to make sure that any exception that's caught is reported, translated accurately, or re-raised.

NathanReb:
Fixes #62 

I've been looking through the code to find other cases than the one mentioned in #61, mostly grep-ing `with` and `exception`.

It all seems good. I've noted two things that could eventually be swallowing exceptions but didn't want to take actions as they probably don't require one:
1. There are a bunch of `try ... with _ ->` which all surround single standard library function calls such as `int_of_string` or `List.nth`. It could eventually be narrowed to catching a specific exception as the doc specifies which one it's supposed to raise. The only thing that I imagine could go wrong is that it could catch a `Sys.Break` if enabled.
2. In [`read.mll`](https://github.com/mjambon/yojson/blob/master/lib/read.mll#L1151) if the finalizer passed to `stream_from_lexbuf` raises in the error case, the exception is silenced. Since the finalizer argument is exposed through the API that's entirely possible. It might be worth logging the exception somehow while still re-raising the one raised by `from_lexbuf`.

If you have anything in mind that I could've missed I'm happy to take a look!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,3 +4,4 @@
 ### Changes
 
 - reraise exceptions in `finish_string` instead of silencing them by raising a `Failure _`
+- raise finalizer exceptions in `from_channel` and `from_lexbuf` readers
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,2 +1,6 @@
 1.4.2 (next)
 ------------
+
+### Changes
+
+- reraise exceptions in `finish_string` instead of silencing them by raising a `Failure _`
--- a/lib/read.mli
+++ b/lib/read.mli
@@ -6,8 +6,10 @@ val compact : ?std:bool -> string -> string
   (** Combined parser and printer.
       See [to_string] for the role of the optional [std] argument. *)
 
-
 (** {2 JSON readers} *)
+
+exception Finally of exn * exn
+(** Exception describing a failure in both finalizer and parsing. *)
 
 val from_string :
   ?buf:Bi_outbuf.t ->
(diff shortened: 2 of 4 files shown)


NathanReb:
I didn't add a test as there aren't any but I'd also be happy to do that if you like it's worth it.

NathanReb:
Just wanted to check if you have had the time to give a look at this @mjambon

mjambon:
Sorry about not reacting. It's hard to find time for this even if it's not much work for me. I'd like to hand off the project to https://github.com/ocaml-community as well as other open-source OCaml projects that I once started. I didn't get around to doing this.

mjambon:
We'll have to document in the next release that new exceptions are being raised.

Regarding this in `read.mll`:
```ocaml
        | e ->
            (try fin () with _ -> ());
            raise e
```

I suppose we could just not catch exceptions raised by `fin`, which is a reasonable policy. I don't know if it's the common behavior that people expect from a try-finally construct. In the absence of better insight, I'd go with what's more common or more expected by users. Perhaps then we'd have to call `close_in_noerr` instead of `close_in` where we have currently `let fin () = close_in ic in`. Not sure which is more helpful to the user.

mjambon:
@NathanReb let me know if you want to make further changes or if I should merge this, which is fine by me.

NathanReb:
I'll take a look at the finalizer thing first and get back to you ASAP!

NathanReb:
So I had a look at what `try-finally` constructs in the wild look like:
- `containers` will raise the finalizer's exception, see [here](https://github.com/c-cube/ocaml-containers/blob/master/src/core/CCFun.ml#L45)
- `base`'s `protect` will raise both using a `Finally of exn * exn`, see [here](https://github.com/janestreet/base/blob/master/src/exn.ml#L68)

I think `base`'s solution definitely is the best. What would you think about adding such an exception in `yojson`?

About `1.`, should I leave it as it is or restrict it to specific exceptions ?

NathanReb:
I pushed some changes.

I updated `CHANGES.md` and added the aforementioned `Finally of exn * exn`. Let me know what you think!

mjambon:
Thanks! This looks like it will get the job done. Merging.

PR #66, opened 2018-12-11
Update opam file to OPAM 2 format

Leonidas-from-XIV:
I have a number of maintenance changes and thought that using the OPAM2 format might be a good point to start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
-_build
+/_build/
+/_opam/
 .merlin
-*.install
+/*.install
--- a/yojson.opam
+++ b/yojson.opam
@@ -1,22 +1,32 @@
-opam-version: "1.2"
+opam-version: "2.0"
 maintainer: "martin@mjambon.com"
 authors: ["Martin Jambon"]
-homepage: "https://mjambon.github.io/mjambon2016/yojson-doc/Yojson.html"
-bug-reports: "https://github.com/mjambon/yojson/issues"
-dev-repo: "https://github.com/mjambon/yojson.git"
-available: [ ocaml-version >= "4.02.3"]
+homepage: "https://github.com/ocaml-community/yojson"
+bug-reports: "https://github.com/ocaml-community/yojson/issues"
+dev-repo: "git+https://github.com/ocaml-community/yojson.git"
+doc: "https://ocaml-community.github.io/yojson/"
 build: [
   ["jbuilder" "subst" "-n" name] {pinned}
   ["jbuilder" "build" "-p" name "-j" jobs]
 ]
-
-build-test: [
-  ["jbuilder" "runtest" "-p" name]
-]
-
+run-test: ["jbuilder" "runtest" "-p" name]
 depends: [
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
-_build
+/_build/
+/_opam/
```
pmetzger: What's the intent here? Only to match _build and _opam in the top directory of the project?

yojson.opam
```diff
@@ -1,22 +1,30 @@
...
 depends: [
+  "ocaml" {>= "4.02.3"}
   "jbuilder" {build}
   "cppo" {build}
   "easy-format"
   "biniou" {>= "1.2.0"}
 ]
+synopsis:
+  "Yojson is an optimized parsing and printing library for the JSON format"
+description: """
```
pmetzger: Should a copy of the synopsis be at the start of the description?

.gitignore
pmetzger: BTW, I'd suggest not changing the .gitignore in the same commit...

Leonidas-from-XIV: Exactly, only match if they are directories and only match if they are on the top of the repository. I don't like random (though unlikely) surprises of `_opam` and `_build` missing due to overzealous ignore rules, when there are exactly two cases that I want to match.

In fact the `*.install` rule could also be reformulated to be `/*.install`…

yojson.opam
Leonidas-from-XIV: I don't know. Since there was no `descr` file I just copied the description and synopsis from the already opam2 converted `opam-repository`, trusting it'd be an accurate representation of the library.

https://github.com/ocaml/opam-repository/blob/9cd5c37b861f9ec741883f3889a89653557105b9/packages/yojson/yojson.1.4.1/opam#L18-L28

.gitignore
Leonidas-from-XIV: I've split the `.gitignore` change into its own commit.

yojson.opam
pmetzger: I believe that you want the description to be standalone, so yes, the synopsis text needs to be at the start of the description as well.

Leonidas-from-XIV: Added in ab4115f.

.gitignore
pmetzger: It looks like it is still here?

Leonidas-from-XIV: It is split out in its own commit, as you requested. I've also took the liberty to amend the original commit to include the revised description, d3a501566b3c673bc67c478b1be980d4a705c50e so the `.gitignore` commit is in 8853750d2f745b826801347ebdbd38ce635b94e9.

pmetzger: I meant it should be merged in a distinct PR. (One PR, one concern.)

But it's not that big a deal.

@NathanReb as the new maintainer, are you happy with this? If so, please merge it!

NathanReb: I think splitting the commit is good, browsing the git history makes the intent pretty clear, which is what matters the most. Let's not put too much of a burden on @Leonidas-from-XIV !

PR #67, opened 2018-12-11
Dune & dune-release

Leonidas-from-XIV:
Based on #66.


.gitignore
```diff
@@ -1,3 +1,4 @@
-_build
+/_build/
+/_opam/
```
pmetzger: This probably conflicts with the change from the other PR. I'd probably suggest doing this in a separate commit if it is needed.

Leonidas-from-XIV: Yes, the first commit is the one from #66. I decided to base it on top of that, to avoid creating a branch that will immediately generate conflicts when the other branch is merged so I'd rebase it on top of master if we decide to merge #66 in some way.

pmetzger: Generally it's best to have such PRs be independently commitable if possible.

mjambon:
Looks good to me.

You should probably make the same changes in opam-repository if you haven't already. We should also replace my name/email by @NathanReb. This can be done later. See https://github.com/ocaml-community/yojson/issues/68

Leonidas-from-XIV:
I suggest using `dune-release opam pkg` (or even that option that directly creates a PR to opam-repository) to generate an `opam` file with the description and synopsis from the repo.

yojson.opam
```diff
@@ -1,22 +1,30 @@
...
-dev-repo: "https://github.com/mjambon/yojson.git"
-available: [ ocaml-version >= "4.02.3"]
+homepage: "https://github.com/ocaml-community/yojson"
+bug-reports: "https://github.com/ocaml-community/yojson/issues"
+dev-repo: "git+https://github.com/ocaml-community/yojson.git"
+doc: "https://ocaml-community.github.io/yojson/"
 build: [
-  ["jbuilder" "subst" "-n" name] {pinned}
-  ["jbuilder" "build" "-p" name "-j" jobs]
+  ["dune" "subst" "-n" name] {pinned}
```
avsm: no need for `-n` with `dune subst` so that can be removed
```suggestion
  ["dune" "subst"] {pinned}
```

yojson.opam
```diff
@@ -1,22 +1,30 @@
...
+  "ocaml" {>= "4.02.3"}
+  "dune" {build}
   "cppo" {build}
   "easy-format"
   "biniou" {>= "1.2.0"}
 ]
+synopsis:
+  "Yojson is an optimized parsing and printing library for the JSON format"
+description: """
+It addresses a few shortcomings of json-wheel including 2x speedup,
```
avsm: This is a rather historical description of yojson, which has outlasted the much-missed json-wheel library :-) Perhaps just describe yojson as a performant JSON serialisation library here?

NathanReb: I think changing the description would indeed be good as some of us (myself included) might be too young to even know about `json-wheel`, but I think this can be done seperately. Also these changes actually come from #66!

yojson.opam
Leonidas-from-XIV: Thanks for the suggestion, applied in a542b6a. 👍

yojson.opam
Leonidas-from-XIV: Yes, sorry about that. I've rebased on top of current master so it is less confusing now. I agree that updating description would be good, since I also don't know `json-wheel` 🙂

Leonidas-from-XIV:
I checked the build and it fails because the Mac builds can't find `opam`, so there is an unrelated problem in ocaml-ci-scripts.

examples/dune
```diff
@@ -0,0 +1,8 @@
+(executables
```
NathanReb: Since there's only one of them we can probably use an `executable` stanza here.

Unless I'm missing something of course!

Makefile
```diff
@@ -1,21 +1,30 @@
...
+	@dune build @install @examples
 
+.PHONY: run-examples
+run-examples:
+	dune exec examples/filtering.exe < examples/filtering.json
+
+.PHONY: test
 test:
-	@jbuilder runtest
+	@dune runtest --force
```
NathanReb: The `test` target can be misleading as it currently neither run nor build anything.

I'd really like to write a test suite for `yojson` but in the meantime I think it would be better to remove this. What do you think?

yojson.opam
```diff
@@ -6,13 +6,13 @@ bug-reports: "https://github.com/ocaml-community/yojson/issues"
 dev-repo: "git+https://github.com/ocaml-community/yojson.git"
 doc: "https://ocaml-community.github.io/yojson/"
 build: [
-  ["jbuilder" "subst" "-n" name] {pinned}
-  ["jbuilder" "build" "-p" name "-j" jobs]
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
 ]
-run-test: ["jbuilder" "runtest" "-p" name]
+run-test: ["dune" "runtest" "-p" name]
```
NathanReb: I think we'd better remove this until we have an actual test suite, same as for the `make` target.

NathanReb:
Yeah don't worry about the CI, I'll see what I can do about that!

Makefile
```diff
@@ -1,21 +1,26 @@
...
-	@jbuilder build @install @DEFAULT
+	@dune build @install @examples
 
-test:
-	@jbuilder runtest
+.PHONY: run-examples
+run-examples:
+	dune exec examples/filtering.exe < examples/filtering.json
 
+.PHONY: check
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Oops, I totally missed that too. Removed in bbb576d.

PR #75, opened 2019-01-28
Reduce macOS builds

Leonidas-from-XIV:
As noticed, building on macOS takes forever and getting all the variations to build is 2h+ of waiting. Therefore I decided to exclude all but the newest OCaml for builds.

There is the questions whether it should not be enough to just test 4.02 and 4.07 on the Linux side since versions in between should just work, but these builds happen all faster than a single build on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,18 +12,5 @@ env:
   - OCAML_VERSION=4.05
   - OCAML_VERSION=4.06
   - OCAML_VERSION=4.07
-matrix:
-  exclude:
-    - os: osx
-      env: OCAML_VERSION=4.02 OCAML_SWITCH=4.02.3
-    - os: osx
-      env: OCAML_VERSION=4.03
-    - os: osx
-      env: OCAML_VERSION=4.04
-    - os: osx
-      env: OCAML_VERSION=4.05
-    - os: osx
-      env: OCAML_VERSION=4.06
 os:
   - linux
-  - osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,18 @@ env:
   - OCAML_VERSION=4.05
   - OCAML_VERSION=4.06
   - OCAML_VERSION=4.07
+matrix:
+  exclude:
+    - os: osx
+      env: OCAML_VERSION=4.02 OCAML_SWITCH=4.02.3
+    - os: osx
+      env: OCAML_VERSION=4.03
+    - os: osx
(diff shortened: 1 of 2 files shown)


Leonidas-from-XIV:
It still takes forever to even get a macOS machine to build (it is Travis' free tier so fair enough), so I suggest dropping macOS builds altogether. It is not like this package is going to work much different on macOS than on GNU/Linux anyway.

.travis.yml
```diff
@@ -12,6 +12,18 @@ env:
   - OCAML_VERSION=4.05
   - OCAML_VERSION=4.06
   - OCAML_VERSION=4.07
+matrix:
+  exclude:
```
NathanReb: Given the builds we want to run I think the include syntax would be a better a fit as it would be more concise. I'm not deeply familiar with the fine grained tweaks of travis config but would the following work?
```yaml
env:
  global:
  - PACKAGE=yojson

matrix:
  include:
  - os: linux
    env: OCAML_VERSION=4.02 OCAML_SWITCH=4.02.3
  - os: linux
    env: OCAML_VERSION=4.03
  - os: linux
    env: OCAML_VERSION=4.04
  - os: linux
    env: OCAML_VERSION=4.05
  - os: linux
    env: OCAML_VERSION=4.06
  - os: linux
    env: OCAML_VERSION=4.07
  - os: osx
    env: OCAML_VERSION=4.07
```

Couldn't find the answer in their documentation unfortunately. Would you mind trying that?

Leonidas-from-XIV:
I was so frustrated because while even when the osx builds run, it still takes about an hour of waiting to get an osx build to start (since they have to run it on real hardware and can't just set up an arbitrary amount of them), so how about this: just remove builds on OS X. Since there is no platform-dependent code I think for the sake of getting quick feedback I feel like sacrificing these builds might be a worthwhile sacrifice.

What do you think?

NathanReb:
I think it's a good idea, please go ahead and remove them!

Leonidas-from-XIV:
Okay, I pushed the change. Some of them are [red](https://travis-ci.org/ocaml-community/yojson/builds/486911964) but for unrelated purposes (PPA down or so?), I believe re-running might solve the issues.

But you can see that Travis stopped to schedule osx builds.

Leonidas-from-XIV:
Yep, rerunning did the trick.

NathanReb:
Restarted the builds to make sure everything was ok.

TYVM!

PR #76, opened 2019-01-28
Revive benchmark

Leonidas-from-XIV:
Since we are not benchmarking against JSON Wheel anymore I took the liberty to remove the dependency and just port the previous benchmark into `Core_bench` which is very good at reporting performance numbers. Since this is not part of what people will need to use yojson or even run its tests I thought I'd prefer to just use an existing mature benchmark solution instead of rolling my own.

Maybe some more demanding benchmark data would be nice, and some more operations to test but for now I'll just await your input.


Leonidas-from-XIV:
For the record, current values on the machine I'm developing on:

```
Estimated testing time 20s (2 benchmarks x 10s). Change using -quota SECS.
┌──────────────┬──────────┬───────────┬──────────┬──────────┬────────────┐
│ Name         │ Time/Run │   mWd/Run │ mjWd/Run │ Prom/Run │ Percentage │
├──────────────┼──────────┼───────────┼──────────┼──────────┼────────────┤
│ JSON reading │  18.23us │ 1_167.02w │    2.23w │    2.23w │    100.00% │
│ JSON writing │  10.45us │   721.00w │  258.30w │    0.30w │     57.33% │
└──────────────┴──────────┴───────────┴──────────┴──────────┴────────────┘
```

PR #82, opened 2019-05-05
Revert deprecation warning on `json` type aliases

NathanReb:
This has been causing troubles all around with no actual benefit.


avsm:
Does this mean that they're not going to be deprecated then? :-)  If so, a release of Yojson master to opam would be appreciated, as the released version still emits them.

PR #85, opened 2019-09-25
Avoid full copy of buffer in reading json

skcho:
`map_string` copies the full buffer (`b.Bi_outbuf.o_s`) with calling `Bytes.to_string`.

* The size of buffer seems to increase consistently during the reading, which makes the copies expensive, especially when there has been a long-length field data. 
* But the function `f` is using only a substring of the buffer.

This PR tries to copy less in `map_string` with `Bytes.sub_string`.

### DIFF
--- a/lib/read.mll
+++ b/lib/read.mll
@@ -317,7 +317,7 @@ and finish_string v = parse
 
 and map_string v f = parse
     '"'           { let b = v.buf in
-                    f (Bytes.to_string b.Bi_outbuf.o_s) 0 b.Bi_outbuf.o_len }
+                    f (Bytes.sub_string b.Bi_outbuf.o_s 0 b.Bi_outbuf.o_len) 0 b.Bi_outbuf.o_len }
   | '\\'          { finish_escaped_char v lexbuf;
                     map_string v f lexbuf }
   | [^ '"' '\\']+ { add_lexeme v.buf lexbuf;


skcho:
> But the function `f` is using only a substring of the buffer.

Please let me know, if this thought is wrong.

******

A similar issue is also in `map_lexeme`.

https://github.com/ocaml-community/yojson/blob/6f8ed3f738aa2132c4c298a3b09ac68512ef8975/lib/read.mll#L152-L154

I believe the last line can be replaced with

```ocaml
f (Bytes.sub_string lexbuf.lex_buffer lexbuf.lex_start_pos len) 0 len
```

but I am not sure what is the assumption about the mapping function `f`.

If you agree with this, I will add this change in this PR too.

skcho:
@NathanReb Could you check this PR out please?

NathanReb:
I was off for a few weeks, I'll take a look ASAP!

rgrinberg:
@Leonidas-from-XIV mind taking a look at this?

Leonidas-from-XIV:
I looked a little bit at this and of course the PR doesn't apply cleanly anymore because we removed `biniou`, but can easily be fixed with `Buffer.sub`, but it is also not really necessary since the current code does `Buffer.to_string` which returns the current content of the buffer, not the underlying length of the buffer.

But in general this code does not seem to be necessary since `map_string` is only used by `map_ident` and that is not used by… anything, so it also isn't called as part of reading JSON.

That said, `map_string` and `map_ident` are part of the undocumented API and I am not entirely sure what they are used for in the first place since this is pretty much part of of the lexing and does not make that much sense outside of it. I suggest just deleting them altogether. @NathanReb if you agree I'll make a PR dropping this.

skcho:
Hello @Leonidas-from-XIV . Thank you for looking into this.

> That said, map_string and map_ident are part of the undocumented API and I am not entirely sure what they are used for in the first place since this is pretty much part of of the lexing and does not make that much sense outside of it.

We were using atdgen to transfer soma data from C++ to OCaml. In the atdgen-generated code, there were `map_ident` calls. I don't know much about recent changes on yojson and atd though.

> but it is also not really necessary since the current code does Buffer.to_string

OK, `map_string` seems to be fixed already. Great! Feel free to close the PR, if you don't mind `map_lexeme` that calls `Bytes.to_string` too.

Leonidas-from-XIV:
> In the atdgen-generated code, there were map_ident calls. I don't know much about recent changes on yojson and atd though.

Ahh, okay, that's what it is for. Interesting. I assume this stays the same but you uncovered a potential incompatibility in the 2.0 release, where this is changed to always return `0` and `length` (to the point of being somewhat silly).

> OK, `map_string` seems to be fixed already. Great! Feel free to close the PR, if you don't mind `map_lexeme` that calls `Bytes.to_string` too.

I've opened #108 to at least align the behaviour of the functions so they're at least consistent and of course slightly more efficient.

skcho:
> and of course slightly more efficient.

As far as I can remember, in `map_string`, the buffer size of `b.Bi_outbuf.o_s` monotonely increased whenever it read very long string inputs. So, after the buffer size became very big, it took so long time to copy the full buffer to string. I am not sure whether there is a similar issue in `map_lexeme` or not though.

> in the 2.0 release

I can't wait it! Thank you.

PR #87, opened 2019-11-14
Discussion / Draft: Support trailing commas for JSON parsing

bryphe:
Hi,

Thanks for all the work on `yojson! I've been using it in a ton of projects - it's great 🎉 

I'm using it to parse user-crafted configuration files, and in those cases, I'd like more 'resiliency' - forgiving of comments and trailing commas. A use case for this is here: https://github.com/onivim/oni2/issues/836

I know this is likely the implementation is likely incorrect - it looks like the preference for this was to fulfill the RFC 8259 spec: https://github.com/onivim/oni2/issues/836 and this change breaks some of the currently-passing RFC 8259 compliances tests, as the RFC doesn't support trailing commas. 

Based on the details in #81 , it sounds like there is still some thought about where a resilient-parser like this would belong (ie, `Yojson.Standard.t` for RFC 8259 standards compliant parsing, and some other module - maybe `Yojson.Safe.t` or `Yojson.Resilient.t`?)

In any case, I wanted to put it up here as a draft in case there was anyone else looking for this-

### DIFF
--- a/lib/read.mll
+++ b/lib/read.mll
@@ -231,20 +231,16 @@ rule read_json v = parse
                    try
                      read_space v lexbuf;
                      read_object_end lexbuf;
-                     let field_name = read_ident v lexbuf in
-                     read_space v lexbuf;
-                     read_colon v lexbuf;
-                     read_space v lexbuf;
-                     acc := (field_name, read_json v lexbuf) :: !acc;
                      while true do
-                       read_space v lexbuf;
-                       read_object_sep v lexbuf;
-                       read_space v lexbuf;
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV:
We had a discussion on this and there was also a proposal to support the complete [JSON5](https://json5.org/) specification, which, along with some other minor quality of life improvements, also brings support for trailing commas.

bryphe:
@Leonidas-from-XIV - thanks for the update! That sounds great.

Leonidas-from-XIV:
Fixed by #152.

Issue #88, opened 2019-11-20
remove dep on easy-format


c-cube:

Leonidas-from-XIV:
Closing this as the PR has been merged, it is slated for the upcoming 2.0.0 release. Thank you @c-cube!

PR #90, opened 2019-11-26
remove dep on easy-format

c-cube:
see #88


lib/pretty.ml
```diff
@@ -1,69 +1,63 @@
...
     | `Intlit s
     | `Floatlit s
-    | `Stringlit s -> Easy_format.Atom (s, Easy_format.atom)
-    | `List [] -> Easy_format.Atom ("[]", Easy_format.atom)
-    | `List l -> Easy_format.List (("[", ",", "]", array), List.map (format std) l)
-    | `Assoc [] -> Easy_format.Atom ("{}", Easy_format.atom)
-    | `Assoc l -> Easy_format.List (("{", ",", "}", record), List.map (format_field std) l)
+    | `Stringlit s -> Format.pp_print_string out s
+    | `List [] -> Format.pp_print_string out "[]"
+    | `List l -> Format.fprintf out "[@;<1 0>@[<hov>%a@]@;<1 -2>]" (pp_list "," (format std)) l
```
Leonidas-from-XIV, line 22: I am a bit confused why the indentation adds 0 in the beginning but removes 2 in the end?

On a "less of a question, more of a comment"-front, I would expect JSON to be formatted `[1, 2, 3]`, that is, without the spaces, but I just checked the old implementation and it does the exact same thing (`[ 1, 2, 3 ]`) so it makes sense to match what it was doing.

lib/pretty.ml
```diff
@@ -1,69 +1,63 @@
...
         if std then
-          format std (`List l)
+          format std out (`List l)
         else
           if l = [] then
-            Easy_format.Atom ("()", Easy_format.atom)
+            Format.pp_print_string out "()"
           else
-            Easy_format.List (("(", ",", ")", tuple), List.map (format std) l)
+            Format.fprintf out "(@,%a@;<0 -2>)" (pp_list "," (format std)) l
```
Leonidas-from-XIV, line 33: I am wondering whether there should be a break-hint after the first `(`? Or is `@,` equivalent to `@;<0 0>`?

lib/pretty.ml
```diff
@@ -1,69 +1,63 @@
...
 
     | `Variant (s, Some x) ->
         if std then
-          format std (`List [ `String s; x ])
+          format std out (`List [ `String s; x ])
         else
-          let op = "<" ^ json_string_of_string s ^ ":" in
-          Easy_format.List ((op, "", ">", variant), [format std x])
+          let op = json_string_of_string s in
+          Format.fprintf out "<@[<hv2>%s: %a@]>" op (format std) x
```
Leonidas-from-XIV, line 46: It took me a while to understand that `<hv2>` probably means `<hv 2>`, correct? Maybe adding the space would alleviate the confusion for future readers?

PR #91, opened 2019-12-05
Add an example for constructing json.

gs0510:


gs0510:
@NathanReb  Can you take a look? :)

examples/constructing.ml
```diff
@@ -0,0 +1,34 @@
...
+      ( "pages",
+        `Assoc
+          [ ("id", `Int 1); ("title", `String "The Art of Flipping Coins") ] );
+    ]
+
+let main () =
+  let oc = stdout in
+  Yojson.Basic.pretty_to_channel oc json_output;
+  output_string oc "\n";
+  close_out oc
```
NathanReb: I think you can get rid of the `close_out` here, it's unnecessary!

PR #100, opened 2020-07-21
Remove `json` type

### LINKED ISSUE #97: Remove json type aliases in 2.0.0
They were deprecated (not in the most subtle way I admit) when we added the `t` type to replace, we should remember to remove them before the major release!

Leonidas-from-XIV:
I've also chose to remove `json_max`. A quick search on GitHub has not shown any usage of it (mind you, the search is very bad, so I possibly missed something).

Closes #97.

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -23,7 +23,6 @@
    write.mli
    read.mli
    safe.mli
-   pretty.mli
    write2.mli
    common.mli
    util.mli
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
   * `to_outbuf` becomes `to_buffer` and `stream_to_outbuf` becomes
     `stream_to_buffer`
 - Removed `yojson-biniou` library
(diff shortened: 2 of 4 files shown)


lib/dune
```diff
@@ -23,7 +23,6 @@
    write.mli
    read.mli
    safe.mli
-   pretty.mli
```
NathanReb, line 26: Any reason why this is removed?

Leonidas-from-XIV: This is very much a deliberate choice. That file is never actually seen by the compiler since it is not included in `yojson.cppo.mli` since the compiler only ever sees `write2.mli` and `pretty.mli` is not part of the public interface.

To wit, check out the commit right before and notice the code compiling just fine despite `pretty.mli` referencing a `json` type that was removed.

I think the whole CPPO magic is extremely intransparent and should be simplified so I opted to at least remove parts that don't seem to have any function to decrease confusion.

NathanReb: Ah my bad! This is unfortunate indeed, let's remove it!

PR #102, opened 2020-07-21
Update README

Leonidas-from-XIV:
This modernizes the `README` a tiny bit to provide a simple example as starting point as well as pointers to related libraries/tools. Can surely be improved but I think it goes in the right direction. Having the API docs published is also great.

I am not 100% sure about the license bit, maybe we can ask @mjambon whether we can change the wording to match the 3-clause BSD license as it is published online exactly? I am sure packagers for distributions would appreciate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,3 +75,10 @@ Check out in particular
 [good first time issues](https://github.com/ocaml-community/yojson/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+time+issue%22)
 and other issues with which
 [we could use some help](https://github.com/ocaml-community/yojson/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22).
+
+
+License
+-------
+
+`Yojson` is licensed under the 3-clause BSD license, see `LICENSE.md` for
+details.
--- a/README.md
+++ b/README.md
@@ -17,15 +17,17 @@ Currently at https://ocaml-community.github.io/yojson/
 Examples
 --------
 
-A simple example on how to parse JSON that is in a string.
+A simple example on how to parse JSON from a string literal.
 
 ```ocaml
-let to_parse = {|
+let json_string = {|
   {"number" : 42,
    "string" : "yes",
    "list": ["for", "sure", 42]}|}
+(* val json_string : string *)
 
-let json = Yojson.Safe.from_string to_parse
+let json = Yojson.Safe.from_string json_string
+(* val json : Yojson.Safe.t *)
 
 Format.printf "Parsed to %a" Yojson.Safe.pp json
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,50 +1,86 @@
-Yojson: low-level JSON library for OCaml
-========================================
+Yojson: JSON library for OCaml
+==============================
 
 [![Build Status](https://travis-ci.org/ocaml-community/yojson.svg?branch=master)](https://travis-ci.org/ocaml-community/yojson)
 
-_This library is for manipulating the json AST directly. For mapping between OCaml types and json, we recommend [atdgen](https://github.com/ahrefs/atd)._
+This library parses complete JSON files into a nested OCaml data structure,
+similarly to DOM parsers for XML. For stream processing in a SAX-like fashion
```
NathanReb, line 7: I'm personally not familiar with SAX so maybe a link would help people like me!

NathanReb: Also Yojson supports parsing streams of JSON values. It's two different things but it might be worth clarifying this to avoid scaring of users. I'm wondering if the sentence `parses complete JSON files into a nested OCaml data structure` doesn't narrow the scope of the library a bit too much.

README.md
```diff
@@ -1,50 +1,86 @@
...
+A simple example on how to parse JSON from a string literal.
+
+```ocaml
+let json_string = {|
+  {"number" : 42,
+   "string" : "yes",
+   "list": ["for", "sure", 42]}|}
+(* val json_string : string *)
+
+let json = Yojson.Safe.from_string json_string
```
NathanReb, line 29: I wonder if we shouldn't advocate the use of `Basic` rather than safe as it represents actual JSON. That will make it easier to drop the other types if everybody starts using the "right" one.

Leonidas-from-XIV: I think the least invasive way to go forward is to remove tuples and variants from `Safe` since that's the variant that nearly all users (including all PPX rewriters) have agreed upon.

But for now I would suggest `Safe` as the "main" mode that we advertise, just to reduce confusion and avoid people running into a problem where they have `Yojson.Basic.t` in their code but get `Yojson.Safe.t` from other libraries.

NathanReb: Yeah that's a good point, especially for derivers! Let's keep it that way!

README.md
Leonidas-from-XIV: I've reformulated it a bit and moved the mention of `jsonm` to the "related" section; you're right there is no point in dragging XML concepts into JSON.

Leonidas-from-XIV:
@mjambon Can we get your input on the license wording?

mjambon:
@Leonidas-from-XIV yes, the version of the license at https://opensource.org/licenses/BSD-3-Clause is clearer on the non-endorsement clause (point 3). This is a minor change and I approve it.

PR #107, opened 2020-10-23
upate opam file to add doc target

c-cube:


Leonidas-from-XIV:
Looks good to me and the Travis failure looks like a transient error, so I'm merging it. Thanks @c-cube!

In the future we should consider letting dune just autogenerate the `opam` file, so it should automatically be up to the current OPAM trends.

PR #108, opened 2020-11-02
Avoid full buffer copy in `map_lexeme`

Leonidas-from-XIV:
As mentioned in #85 by @skcho, the code in `map_ident`/`map_string` was changed to not copy the whole buffer and give a start/end point (though this happened somewhat more on the accidental than deliberate side) thus subtly changing the semantics of the data `f` was called on. This saves copying but prevents `f` from peeking beyond the borders it was
assigned. Whether this breaks consumers depends on what they do with it.

Nevertheless, what this PR does is it applies the same logic to `map_lexeme`, so while the semantics might have changed at least they change in a consistent way.

Closes #85

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,11 @@
 - The function `to_file` now adds a newline at the end of the generated file. An
   optional argument allows to return to the original behaviour (#124, @panglesd)
 
+### Fix
+
+- Avoid copying unnecessarily large amounts of strings when parsing (#85, #108,
+  @Leonidas-from-XIV)
+
 ## 1.7.0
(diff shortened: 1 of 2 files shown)


panglesd (approved):
I'm not really sure about this one.
I checked in the `atd` codebase, and `map_lexeme` does not seem to be used (anyway, `atd` will have to change quite a few things as yojson does not use `outbuf` anymore). 
I search for `map_lexeme` on github and it does not seem to be anywhere except old vendored libraires.

So I think changing it to be more consistent with `map_ident`/`map_string` is ok!

PR #116, opened 2021-11-22
Run changelog action against `master` branch

Leonidas-from-XIV:
This is currently the branch that this repo uses as default.


NathanReb:
Argh good catch! We could also rename the branch!

NathanReb (approved):

Leonidas-from-XIV:
@NathanReb Good idea but I would wait for this after we get 2.0 out of the door to start off some major janitoring work (like removing CPPO and the silly variants) for 3.0.

PR #117, opened 2021-11-30
Add bench dependencies on core and core_bench.

tmcgilchrist:


tmcgilchrist:
There's one failure on 4.02 with compiling `core_extended` that's a dependency of `core_bench`.
This [version](https://opam.ocaml.org/packages/core_extended/core_extended.113.33.03/) is the last to support 4.02  which seems to have the bug https://github.com/janestreet/core_extended/issues/12.
Putting a lower bound of 4.03 on the opam file would avoid this issue, not sure if anyone would be actively using 4.02 anymore?

c-cube:
benchmarks don't seem to be run as part of `opam install yojson -t`, so I don't think this should be a dependency. If people want to benchmark yojson locally they can install the dependencies.

Leonidas-from-XIV:
Yes, the benchmark is not part of the test, partly because that would be a rather heavy dependency and partly because they don't really test anything as there is no benchmark value so it wouldn't be possible to fail or succeed a test. It would be great if we could specify a `with-bench` variable in `opam` but custom variable handling in OPAM is somewhat arcane.

What would however be necessary is to figure out how to exclude `dune` from building this code on CI because it is failing the build on the `master` branch. I thought that moving it into its own `@bench` alias would do so but apparently it is still part of `@install @check @runtest`

(Overall it would be nice to run this code on CI to notice bitrot, but we'd need to have a way to specify independent test steps)

tmcgilchrist:
`ocaml-ci` is driven off the `opam` files it finds in the repository to work out what to build. Something like `with-bench` or another way of introducing different classes of tests/benchmarks would be very useful but isn't currently available.

So one option is to introduce a second opam file `yojson-bench.json`, have that depend on yojson/core_bench, and change the opam build config to build the benchmarks but not run them in CI.  https://github.com/ocaml-community/yojson/pull/117/commits/bd0d4fd1d08502c09048db6dd4d1d46a53c568a1

Leonidas-from-XIV:
I'm wondering whether this is a changelog-less change. Given the previous binary wasn't ever installed, it is not much of a change for users going forward. But they can install the binary now (though it feels somewhat nonsensical to submit `yojson_bench` to opam-repository, given it only exists to run in CI)

tmcgilchrist:
Agree that submitting `yojson_bench` to opam would be a bad idea. 

There's an unfortunate functionality gap between needing an opam file to drive CI and not wanting something as a package when submitting upstream via `dune-release`.

I don't have a good solution for that, perhaps the best option is to edit the opam-repository PR and remove the mention of `yojson_bench`. Having that as an option in `dune-release` would be useful.

Leonidas-from-XIV:
@tmcgilchrist Could you add a short entry in the changelog about this? I think we should merge it, for the CI to be helpful again and then we could see if we find a better solution for this later. Let's not have perfect be the enemy of good.

panglesd (approved):
Looks good to me :)

PR #119, opened 2022-01-17
Add package to build the benchmark

Leonidas-from-XIV:
This is probably not the best approach to deal with the benchmarks that have heavy dependencies but possibly a starting point on how to get it sorted out.


Leonidas-from-XIV:
Already handled by #117 with essentially the exact same approach, closing.

PR #122, opened 2022-01-21
remove restriction against generating JSON with an atomic root value

### LINKED ISSUE #121: `pretty_print` raises errors when asked to print any value other than an array or object
e.g.:

```
utop # Format.printf "%a@." (Yojson.Basic.pretty_print ~std:true) (`Int 5);;
Exception:
Yojson.Json_error
 "Root is not an object or array as requested by the JSON standard"
```
My first question is, where does the standard request that JSON values aside from objects and arrays not be printed? Looking at [json.org](http://json.org), and the actual [ECMA standard](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf), I don't see any language to this effect. On the contrary, the ECMA document states plainly that "A JSON value can be an object, array, number, string, true, false, or null", and the quasi-BNF on the side of json.org is pretty clear that any of these things is a valid JSON value.

Beyond litigating the specification though, the current behaviour is just incredibly unhelpful. Javascript runtimes happily `parse` and `stringify` non-object and non-array "root" values, and OCaml programs need to deal with the same. To that end, I consistently carry around these functions so I can serialize JSON values as required by APIs and other programs (with the understanding that my treatment of strings via Format's `%S` might actually be slightly off):

```ocaml
let pp_json f fmt =
  let open Format in
  function
  | `Null -> pp_print_string fmt "null"
  | `Int v -> pp_print_int fmt v
  | `Float v -> pp_print_float fmt v
  | `String v -> fprintf fmt "%S" v
  | `Bool v -> pp_print_bool fmt v
  | v -> fprintf fmt "@[%a@]" f v

let pp_json_basic = pp_json (Yojson.Basic.pretty_print ~std:true)

let pp_json_safe = pp_json (Yojson.Safe.pretty_print ~std:true)
```

Can we remove this limitation?

cemerick:
This fixes #121.

Unfortunately no tests to add; as it stands, there's a only a single monolithic test for pretty-printing, and the simple change doesn't seem to warrant setting up a new harness.

### DIFF
--- a/lib/pretty.ml
+++ b/lib/pretty.ml
@@ -49,11 +49,7 @@ and format_field std out (name, x) =
   Format.fprintf out "@[<hv2>%s: %a@]" (json_string_of_string name) (format std) x
 
 let pp ?(std = false) out x =
-  if std && not (is_object_or_array x) then
-    json_error
-      "Root is not an object or array as requested by the JSON standard"
-  else
-    Format.fprintf out "@[<hv2>%a@]" (format std) (x :> t)
+  Format.fprintf out "@[<hv2>%a@]" (format std) (x :> t)
 
 let to_string ?std x =
   Format.asprintf "%a" (pp ?std) x
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@
 - Removed `yojson-biniou` library
 - Removed deprecated `json` type aliasing type `t` which has been available
   since 1.6.0 (@Leonidas-from-XIV, #100).
+- Removed constraint that the "root" value being rendered (via either
+  `pretty_print` or `to_string`) must be an object or array. (@cemerick, #121)
 
 ### Add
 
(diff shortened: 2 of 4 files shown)


Leonidas-from-XIV:
Could you add a note in the changelog? Not that I expect things to be breaking a lot but it would probably be helpful to users to know since when the behavior was improved.

cemerick:
Sure, my bad, just updated to include a changelog entry.

lib/pretty.ml
```diff
@@ -49,11 +49,7 @@ and format_field std out (name, x) =
   Format.fprintf out "@[<hv2>%s: %a@]" (json_string_of_string name) (format std) x
 
 let pp ?(std = false) out x =
-  if std && not (is_object_or_array x) then
```
Leonidas-from-XIV, line 52: It looks like `is_object_or_array` is not used anymore (since these were the only uses of it and I can't imagine where else this would be useful), could you remove it too? Avoids us dragging around dead code.

Leonidas-from-XIV (changes requested):
Looks good, I think we can remove even more dead code.

We just fixed the CI (finally), could you rebase on current master?

lib/pretty.ml
cemerick: Sure, done :+1:

Leonidas-from-XIV (approved):

PR #126, opened 2022-01-26
Switch badge to point to OCaml CI

Leonidas-from-XIV:
The code is inspired by the identical badge from dune-release, adapting for the URLs and branch.


panglesd (approved):
Nice!

Issue #54, opened 2018-03-12
Way to update Yojson.Basic.json member?


sgrove:
If I have a json object, is there a function that takes a member, a function of `Yojson.Basic.member -> Yojson.Basic.member`, and returns a new json object with an updated value at the key?

I'm looking for something similar to OCaml's `with`, specifically because I need to update a list of a key to shorten it to a sublist, then pass the whole structure to another function.

mjambon:
No, there isn't such thing. We could add it, but we'd also need to add similar functions for mutating data, with consistent naming. Hopefully we won't end up embedding a full javascript implementation.

seanpoulter:
I'm brand new to OCaml but would be willing to have a try at this. Do you know any examples of these update functions @mjambon? I've found [one StackOverflow post](https://stackoverflow.com/questions/37250474/how-to-update-json-object-using-yojson) so far.

seanpoulter:
I've been hacking away on this tonight to transfer over the Map.update syntax to update a JSON object as an `Assoc (string : json) list. Any thoughts? Is this anywhere near what you're expecting?
```ml
let update key f xs =
    let keep_replace_or_remove acc (name, value) =
        if key = name then
            match f value with
            | None -> acc
            | Some new_value ->
                if value == new_value then acc @ [(name, -1)]
                else acc @ [(name, new_value)];
        else acc @ [(name, value)];
    in

    match xs with
    | `Assoc xs -> `Assoc (List.fold_left keep_replace_or_remove [] xs)
    | _ -> xs
;;


(* Add *)
update "a" (fun a -> Some (a + 98)) (`Assoc [("a",1);("b",2);("c",3);("d",4)]);;
(* - : [> `Assoc of (string * int) list ] = `Assoc [("a", 99); ("b", 2); ("c", 3); ("d", 4)]*)

(* Remove *)
update "a" (fun a -> None) (`Assoc [("a",1);("b",2);("c",3);("d",4)]);;
(* - : [> `Assoc of (string * int) list ] = `Assoc [("b", 2); ("c", 3); ("d", 4)]*)
```

From what I've seen I think a recursive call is preferred over the fold so it probably needs more work.

seanpoulter:
I've made some progress! 🎉

[...]

PR #127, opened 2022-01-26
Add few functions to the utils module

panglesd:
I'm not sure whether this is useful, or if the api is well designed, but it has been requested to add some more function to the Utils module in #54.

What do you think :question:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,8 @@
 
 - Add an opam package `yojson-bench` to deal with benchmarks dependency
   (@tmcgilchrist, #117)
+- Added utilitaries function `assoc_map`, `assoc_map_filter`, `add`, `update`
+  and `remove` to the `Utils` module (@panglesd, #127)
 
 ### Change
 
--- a/lib/util.ml
+++ b/lib/util.ml
@@ -202,3 +202,49 @@ let combine (first : t) (second : t) =
   match (first, second) with
   | (`Assoc a, `Assoc b) -> (`Assoc (a @ b) :  t)
   | (a, b) -> raise (Invalid_argument "Expected two objects, check inputs")
+
+let assoc_map f json = match json with
+  | `Assoc l -> 
+    let rec map res l = match l with
+        [] -> res
+      | (k,v) :: q -> map ( (k,f (k,v)) :: res) q in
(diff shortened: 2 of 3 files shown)


Leonidas-from-XIV:
Personally I am not much of a fan how partial the functions are - it feels like supplying them is a potential foot-gun to users.

Without these functions users have to match on `` `Assoc`` and get the compiler to warn them if the data is not an `` `Assoc`` (in which case they're also free to throw an exception or return an `Error` `result`), whereas these functions hide the match and just throw an exceptions themselves. Also, if we implement these functions the users have to deal with our implementations of `update`, `filter_map` or `map`.

What I would suggest instead is a function like

```ocaml
let with_assoc ~f = function
  | `Assoc xs -> `Assoc (f xs)
  | otherwise -> otherwise
```

So users can use `Yojson.Util.with_assoc ~f:(List.Assoc.update ...) json` or whatever other function they want to operate on. But even that function feels a bit fragile, because in the failure mode it silently does… nothing.

panglesd:
I agree, I was pretty unsure of the need of such functions. As the type is public everything can be done by the user, in the way they prefer.

I think that your suggestion is good though, except that to make it more consistent with `to_assoc` I would raise an exception in case of not `Assoc` type.
And, I guess that this means that `with_bool`, `with_int`, `with_list`... also have to be implemented?

Leonidas-from-XIV:
Yes, that's somewhat the thing… none of these functions nor their behavior in the non-assoc/non-list/non-bool case is obvious and throwing an exception in the rare case is a great way to get your program to explode with an uncaught exception. We could of course make a function like

```ocaml
let with_list ~f = function
  | `List xs -> Ok (f xs)
  | otherwise -> Error (`Msg "Not a list")
```

but that will require the user to do a pattern match around the call of `with_list` (or, should we choose to use exceptions, an easy to forget try-block) so I think these functions are probably still more annoying than helpful. But that's just my, kinda minimalist, opinion.

panglesd:
Okay, so I close this PR with the following arguments:
- the `Utils` module is not that useful: as the `Safe.t` `Basic.t`, ... are public, the library user can write his own utilities according to his preferences (exceptions, `Result`, unmodified input, ...)
- it is still kept for retro-compatibility, but not expanded as this PR did.

Issue #112, opened 2021-10-06
Too much clicking required to see documentation


johnwhitington:
Just started using Yojson, for which many thanks. I am round-tripping PDFs through JSON.

https://ocaml-community.github.io/yojson/

is mostly empty and points to

https://ocaml-community.github.io/yojson/yojson/index.html

which is mostly empty and points to (for most users)

https://ocaml-community.github.io/yojson/yojson/Yojson/index.html

Is there a way to squish some of this together? Meanwhile I have used a bookmark, of course.

Issue #44, opened 2017-07-24
Homepage of documentation missing several useful submodules


theindigamer:
IMHO, as a new user, it would be great to have a short description (say 1 or 2 sentences) on the documentation homepage of what each module does. Currently there are no entries in the table  for `Basic, `Raw`, `Safe` and `Util`. That way, one can get a quick at-a-glance comparison between the different modules without diving into each one from the `Yojson` page and remembering which one does what.

If you don't want to put this information on the documentation homepage, you could add it to the `Yojson` module page.

Leonidas-from-XIV:
Is this still a problem with the new API documentation generated at https://ocaml-community.github.io/yojson/ ?

Issue #106, opened 2020-07-31
[RFC] Support for JSON5 parsing


Leonidas-from-XIV:
# Current state & Rationale

Yojson currently supports JSON parsing as defined in RFC 8259 with a couple of extensions: variants, tuples and comments (the latter not documented).

JSON is a format that is derived from JavaScript and since JSON was created a lot of complaints have been raised that the format is suboptimal in some ways and could be improved, amongst these are

 * Comments support (Yojson already supports this)
 * Trailing commas (there is a draft PR for this, #87)
 * Small details like bare keys

@c-cube mentioned that a lot of these assorted sets of improvements were bundled together into a superset of JSON called [JSON5](https://json5.org/).

# Proposal

This RFC suggests implementing a parser for JSON5. The advantage of implementing JSON5 is that this format is basically updating JSON to JSON as it is specified in the ES5, thus a commonly accepted standard and not just a loose set of extensions to JSON.

The parser could work completely transparently and just return one of the existing JSON variants, like `Safe.t` since JSON5 can be represented in standard-JSON as well. As such they would use the same pretty-printer. It would also avoid creating yet another variant.

This is all in accordance to Postel's law as Yojson would be able to accept JSON5 while being conservative and only output widely-accepted JSON.

# Impact

Depending on how it is decided Yojson could become a pure JSON5 parser, thus all the variants would opt in to use the JSON5 parser, so some non-standard JSON that was not parseable before could become parseable. Alternatively, an opt-in approach is also possible where strict RFC 8259 compliance is necessary.

# Roadmap

Could happen any time after 2.0.0, though a decision on #104 would be useful to know whether these extensions would be part of a JSON5 parser or not.

dhilst:
Hello, I would like to help. I guess we want to extend the current parser to support JSON5 right? I will take a look at the code

c-cube:
yeah, I think the idea would be to have a separate AST (and parser?) for JSON5 :). The AST is richer because there's things like datetimes, if I recall correctly.

dhilst:
Oh okay, I will try it and open a PR when get something

dhilst:
I see that @issuu started to work on this, should I get from where issuu stopped?

mjambon:
I don't think we necessarily need a new AST type for JSON5. The short example showing the JSON5 extensions from https://json5.org/ is:
```
{
  // comments
  unquoted: 'and you can quote me on that',
  singleQuotes: 'I can use "double quotes" here',
[...]

Issue #105, opened 2020-07-31
[RFC] Reduce the amount of different Yojson variants


Leonidas-from-XIV:
# Current state & Rationale

Yojson currently supports parsing into three different variants of JSON: `Basic` (which is mostly "normal JSON"), `Safe` (JSON extended with variants and tuples and making sure numbers don't get truncated; what most Yojson users seem to have settled on) and `Raw` (which parses numbers into strings and does not do string quoting).

This creates friction when different codebases use different types: library A has a hard time interoperating with library B if library A uses `Basic.t` and library B uses `Safe.t`.

# Proposal

When picking Yojson for parsing JSON there is very little guidance on which of these variants to use. When parsing people want to parse JSON correctly into some mostly sensible format and not be confronted with minutia about different ways to parse JSON.

As such this RFC proposes to rethink the way the Yojson variants are divided up. There should be one variant that will fulfill 80% of peoples uses, presumably similar to what `Safe` is today. There should probably also be one that does as little interpretation of values as possible, e.g. never assuming numbers are `` `Int``.

Reducing the amount of variants also makes it easier to remove code-generation via CPPO, which is currently preventing Merlin from working properly on the codebase and making reasoning over the functionality harder by adding a metalevel to the code.

The exact resulting variants are subject to discussion that I want to open up in this RFC.

# Impact

`Yojson.Safe.t` is a type that is used by `ppx_deriving_yojson` and `ppx_yojson_conv` and changing the representation there might cause breakage. The upside is that fixing the PPXes will also automatically fix the users of it.

# Roadmap

This would be planned for a 3.0.0 release at the earliest.

cemerick:
Given the recent release, I wonder if the plan as stated here (tl;dr: re-center on `Safe`, removing `Basic`?) remains the plan?

Leonidas-from-XIV:
I think so, but for the fallout we'd need to check the revdeps and see how much breakage this will cause. After doing a couple of PRs to fix (`json` -> `t`) breakage at least the codebases that I have contributed to use mostly `Safe`, so it might not be too bad.

What would probably be the best is to create one release where `Basic` would be marked as deprecated so people can prepare. However I think at first we should tackle #104 given next to nobody uses these variants and most comments about this are positive.

mjambon:
[...]

PR #128, opened 2022-01-27
Improved the entry point of the documentation.

panglesd:
This PR adds an `index.mld` to improve the entry point of the documentation.

It allows to:
- Hide packages that we do not want to include in the doc (such as `yojson-bench`)
- Describe the library in a short page, without the details of the API.

I think that it solves #112 and #44.

### DIFF
--- a/lib/yojson.cppo.mli
+++ b/lib/yojson.cppo.mli
@@ -3,7 +3,7 @@
 
    - The {{!basic}Basic} JSON type,
    - The {{!safe}Safe} JSON type, a superset of JSON with safer support for integers,
-   - The {{!raw}Raw} JSON type, a superset of JSON, safer but less integrated with ocaml types.
+   - The {{!raw}Raw} JSON type, a superset of JSON, safer but less integrated with OCaml types.
 
 Each of these different types have their own module.
 
--- a/lib/yojson.cppo.mli
+++ b/lib/yojson.cppo.mli
@@ -1,30 +1,19 @@
 (**
-   The Yojson library provides runtime functions for reading and writing JSON
-   data from OCaml. It addresses a few shortcomings of its predecessor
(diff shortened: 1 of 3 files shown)


lib/yojson.cppo.mli
```diff
@@ -1,30 +1,19 @@
...
-   optional quotes around field names, tuples and variants.
-   
-   @author Martin Jambon
-   @see <http://json.org> JSON specification
- *)
+   The Yojson library provides several types for representing JSON values, with different use cases.
+
+   - The {{!basic}Basic} JSON type,
+   - The {{!safe}Safe} JSON type, a superset of JSON with safer support for integers,
+   - The {{!raw}Raw} JSON type, a superset of JSON, safer but less integrated with ocaml types.
```
Leonidas-from-XIV, line 6:
```suggestion
   - The {{!raw}Raw} JSON type, a superset of JSON, safer but less integrated with OCaml types.
```

Leonidas-from-XIV (changes requested):
I think it is nice. While I am aware that most of it is just moving documentation around, I think some of it should be updated to not state things that aren't the case anymore.

panglesd:
I added a mention that one of the goal of yojson is to eventually support the json5 standard... I think that's a reasonable goal to have but I can remove this line if you prefer, as no work has started on this feature!

Leonidas-from-XIV:
@panglesd I have made a proposal in #106 and started an experimental implementation but I am not sure there is a consensus on it yet. I'd suggest we first decide what to do with that RFC :-)

panglesd:
Sure! I removed the json5 mention and implemented your comments.
We'll discuss json5 in #105.

Leonidas-from-XIV (approved):

PR #130, opened 2022-02-09
Add dependency on camlp-streams package for compat

### LINKED ISSUE #129: Deprecation of `Stream`
While we did a good job for the OCaml 5.0 readiness by removing `biniou` and `easy-format` OCaml 4.14 will deprecate the `Stream` module.

We should consider what to do:
  1. Use the `camlp-stream` polyfill that provides this module
  2. Migrate to something else like `Seq`

The issue with 1. is that we're adding another dependency. It's probably not a big problem, but it is still not great to depend on something that was removed from the standard library, for most likely good reasons.

The problem with 2. is that this might decrease the performance and since it breaks the interface, it would need to happen for 2.0.0 (or 3.0.0…).

Leonidas-from-XIV:
Closes #129.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -31,7 +31,7 @@
 - Avoid copying unnecessarily large amounts of strings when parsing (#85, #108,
   @Leonidas-from-XIV)
 - Avoid deprecation warning when building with OCaml 4.14+ by switching to the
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV:
Depends on https://github.com/ocaml/camlp-streams/pull/3 being merged and released to OPAM-Repository.

PR #131, opened 2022-02-09
Stream to Seq

gasche:
Before:

```
┌─────────────────────────────────────┬──────────┬────────────┬────────────┬──────────┬────────────┐
│ Name                                │ Time/Run │    mWd/Run │   mjWd/Run │ Prom/Run │ Percentage │
├─────────────────────────────────────┼──────────┼────────────┼────────────┼──────────┼────────────┤
│ JSON stream roundtrip               │ 998.06us │ 70_051.93w │ 10_985.58w │   13.98w │    100.00% │
└─────────────────────────────────────┴──────────┴────────────┴────────────┴──────────┴────────────┘
```

After:

```
┌────────────────────┬──────────┬────────────┬────────────┬──────────┬────────────┐
│ Name               │ Time/Run │    mWd/Run │   mjWd/Run │ Prom/Run │ Percentage │
├────────────────────┼──────────┼────────────┼────────────┼──────────┼────────────┤
│ JSON seq roundtrip │ 907.16us │ 60_038.93w │ 10_980.64w │    8.75w │    100.00% │
└────────────────────┴──────────┴────────────┴────────────┴──────────┴────────────┘
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,8 @@
 
 - The function `to_file` now adds a newline at the end of the generated file. An
   optional argument allows to return to the original behaviour (#124, @panglesd)
+- The `stream_from_*` and `stream_to_*` functions now use a `Seq.t` instead of a
+  `Stream.t`, and they are renamed into `seq_from_*` and `seq_to_*` (@gasche, #131).
 
 ### Fix
 
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -25,6 +25,13 @@ let large_string_list =
   in
   `List strings
 
+let streamable_string =
+  let buf = Buffer.create (large * 100) in
+  for i = 1 to large do
+    Printf.bprintf buf "%d\n" i
+  done;
+  Buffer.contents buf
+
 let main () =
   Command.run (Bench.make_command [
     Bench.Test.create ~name:"JSON reading" (fun () ->
@@ -46,6 +53,13 @@ let main () =
     Bench.Test.create ~name:"JSON writing assoc to channel" (fun () ->
       Out_channel.with_file "/dev/null" ~f:(fun oc ->
       ignore (Yojson.Safe.to_channel oc large_int_assoc)));
+    begin
+      let buf = Buffer.create 1000 in
+      Bench.Test.create ~name:"JSON stream roundtrip" (fun () ->
+        let stream = Yojson.Safe.stream_from_string ~buf streamable_string in
+        ignore (Yojson.Safe.stream_to_string ~buf stream)
+      )
+    end;
   ])
 
 let () =
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -55,9 +55,9 @@ let main () =
       ignore (Yojson.Safe.to_channel oc large_int_assoc)));
     begin
       let buf = Buffer.create 1000 in
-      Bench.Test.create ~name:"JSON stream roundtrip" (fun () ->
-        let stream = Yojson.Safe.stream_from_string ~buf streamable_string in
-        ignore (Yojson.Safe.stream_to_string ~buf stream)
+      Bench.Test.create ~name:"JSON seq roundtrip" (fun () ->
+        let stream = Yojson.Safe.seq_from_string ~buf streamable_string in
+        ignore (Yojson.Safe.seq_to_string ~buf stream)
       )
     end;
   ])
--- a/lib/read.mll
+++ b/lib/read.mll
@@ -1143,30 +1143,30 @@ and junk = parse
 
   exception Finally of exn * exn
 
-  let stream_from_lexbuf v ?(fin = fun () -> ()) lexbuf =
+  let seq_from_lexbuf v ?(fin = fun () -> ()) lexbuf =
     let stream = Some true in
-    let f i =
-      try Some (from_lexbuf v ?stream lexbuf)
+    let rec f () =
+      try Seq.Cons (from_lexbuf v ?stream lexbuf, f)
       with
           End_of_input ->
             fin ();
-            None
+            Seq.Nil
         | e ->
             (try fin () with fin_e -> raise (Finally (e, fin_e)));
             raise e
     in
-    Stream.from f
+    f
 
-  let stream_from_string ?buf ?fname ?lnum s =
+  let seq_from_string ?buf ?fname ?lnum s =
     let v = init_lexer ?buf ?fname ?lnum () in
-    stream_from_lexbuf v (Lexing.from_string s)
+    seq_from_lexbuf v (Lexing.from_string s)
 
-  let stream_from_channel ?buf ?fin ?fname ?lnum ic =
+  let seq_from_channel ?buf ?fin ?fname ?lnum ic =
     let lexbuf = Lexing.from_channel ic in
     let v = init_lexer ?buf ?fname ?lnum () in
-    stream_from_lexbuf v ?fin lexbuf
+    seq_from_lexbuf v ?fin lexbuf
 
-  let stream_from_file ?buf ?fname ?lnum file =
+  let seq_from_file ?buf ?fname ?lnum file =
     let ic = open_in file in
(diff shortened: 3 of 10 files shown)


gasche:
See #129.

Leonidas-from-XIV:
That's great.

The only problem I see is with stdcompat. After seing how it is built and maintained (autotools, no dune, which creates issues for us using it in the duniverse and is also kinda a large amount of code to depend on) I would much prefer a different solution. Given we infamously use CPPO we could have our own Seq module for older versions of OCaml or something.

gasche:
Actually I think I could use [c-cube/seq](https://github.com/c-cube/seq/) instead, I can update the PR, but only later in the day (or feel free to do it if you want).

Leonidas-from-XIV:
That's a good point, @c-cube's `seq` is exactly the kind of library I hoped for.

CHANGES.md
```diff
@@ -25,6 +25,8 @@
 
 - The function `to_file` now adds a newline at the end of the generated file. An
   optional argument allows to return to the original behaviour (#124, @panglesd)
+- The stream_from_* and stream_to_* functions now use a Seq.t instead of a Stream.t,
```
c-cube, line 28:
```suggestion
- The `stream_from_*` and `stream_to_*` functions now use a Seq.t instead of a Stream.t,
```

Leonidas-from-XIV: Done, also quoted the other identifiers.

bench/bench.ml
```diff
@@ -46,6 +53,13 @@ let main () =
     Bench.Test.create ~name:"JSON writing assoc to channel" (fun () ->
       Out_channel.with_file "/dev/null" ~f:(fun oc ->
       ignore (Yojson.Safe.to_channel oc large_int_assoc)));
+    begin
+      let buf = Buffer.create 1000 in
+      Bench.Test.create ~name:"JSON seq roundtrip" (fun () ->
+        let stream = Yojson.Safe.seq_from_string ~buf streamable_string in
+        ignore (Yojson.Safe.seq_to_string ~buf stream)
+      )
+    end;
```
Leonidas-from-XIV, line 62: Any particular reason for the `begin`/`end`?

lib/read.mll
```diff
@@ -1143,30 +1143,30 @@ and junk = parse
 
   exception Finally of exn * exn
 
-  let stream_from_lexbuf v ?(fin = fun () -> ()) lexbuf =
+  let seq_from_lexbuf v ?(fin = fun () -> ()) lexbuf =
     let stream = Some true in
-    let f i =
-      try Some (from_lexbuf v ?stream lexbuf)
+    let rec f () =
+      try Seq.Cons(from_lexbuf v ?stream lexbuf, f)
```
Leonidas-from-XIV, line 1149:
```suggestion
      try Seq.Cons (from_lexbuf v ?stream lexbuf, f)
```

(low-key wishing we could use automatic code formatting so I wouldn't complain about such petty things).

Leonidas-from-XIV (commented):
Some things I have encountered while going through the code.

bench/bench.ml
gasche: I need some bracketing for the `let buf = ... in ...`, otherwise it would consume following elements of the list I believe. I used begin..end, would you prefer parentheses?

lib/read.mll
gasche: Done!

gasche:
(Note: I'm not working from a flambda switch.)

c-cube:
> (Note: I'm not working from a flambda switch.)

Most people aren't, so I think it's reasonable :grin:

tcoopman:
By coincidence I stumbled on this PR, I was looking into what streaming functions yojson had to offer and because this PR was just opened, I wanted to try it.

Now I have an issue with `Yojson.Basic.seq_to_file "path/to/file" seq`. It runs fine, creates the file, but the file stays empty?

If I run the exact same code but with `Yojson.Basic.seq_to_string` it outputs the correct string

Any idea if I'm doing something wrong, or if something else is going on? 
I'll see if I can find a minimal reproduction

tcoopman:
Here is a minimal (I'm using base) that fails:

```ocaml
let test () =
  let seq = Sequence.of_list [`String "a"; `String "b"; `String "c"] |> Sequence.to_seq in
  Yojson.Basic.seq_to_string seq |> Stdio.print_endline;
  Yojson.Basic.seq_to_file "test.json" seq
```

This prints "a""b""c" on terminal, but "test.json" is empty

gasche:
This does look like a bug indeed, that (according to my test) was already present in the `Stream` functions this PR is replacing. (I guess `seq_to_file` isn't used much!)

lib/write.ml
```diff
@@ -467,34 +467,34 @@ let to_file ?len ?std ?(newline = true) file x =
...
 
-let stream_to_channel ?buf ?(len=2096) ?std oc st =
+let seq_to_channel ?buf ?(len=2096) ?std oc st =
   let ob =
     match buf with
         None -> Buffer.create len
       | Some ob -> ob
   in
-  stream_to_buffer ?std ob st
+  seq_to_buffer ?std ob st
```
tcoopman, line 492: This line is missing a line to actually write the buffer `Buffer.output_buffer oc ob`

tcoopman:
@gasche just found it - see my review - it's missing a line to write the buffer. At least that is how the regular `to_channel` works

lib/write.ml
gasche: Yep, I just found the same fix. I think it should be fixed separately from this PR.

gasche:
Could you try the following implementation?

```ocaml
let seq_to_channel ?buf ?(len=2096) ?std oc st =
  let ob =
    match buf with
        None -> Buffer.create len
      | Some ob -> ob
  in
  Seq.iter (fun v ->
    to_buffer ?std ob v;
    Buffer.output_buffer oc ob;
    Buffer.clear ob;
  ) st
```

tcoopman:
Awesome, that fixes it! :heart:

lib/write.ml
Leonidas-from-XIV: I think it is okay to add this to this PR, but it is up to you.

tcoopman: I think the fix should be the suggestion made by @gasche in the comments, otherwise the `seq` has no real value as there is no memory improvement.

gasche:
The CI failed on <4.07 versions, I think it's because I forgot to add the `seq` dependency to `dune` in addition to `yojson.opam`, I just pushed an updated version.

gasche:
See #133 (fixing the bugs reported by @tcoopman) and #132 (improving the usage of the `?buf` parameter in other `write` functions). #133 conflicts with the current PR, I'll let maintainers decide which one they want to merge (if any) and rebase the other accordingly.

Leonidas-from-XIV:
@gasche #133 is merged, can you rebase please? That would fix our CI too since OCaml-CI added 4.14 already and we're failing on it.

gasche:
Done!

Leonidas-from-XIV:
Looks like `List.to_seq` was only added in 4.07. I've [opened an issue on `seq`](https://github.com/c-cube/seq/issues/11) to see what @c-cube thinks about adding it, but in the meantime we could just implement our own version.

gasche:
I changed the code to avoid `List.seq` and rebased.

Leonidas-from-XIV:
`Seq.cons` was added in 4.11, but given it is just `fun x xs -> Seq.Cons (x, xs)` it is an easy fix to get it to work throughout all OCaml versions.

gasche:
Damn! Sorry for the noise, I did another attempt with just `Seq.Cons` and `Seq.Nil`.

bench/bench.ml
Leonidas-from-XIV: Ah, I looked at it and didn't expect the `;` to bind closer as a statement separator than as a list item separator but I guess thinking about it, it makes sense.

PR #133, opened 2022-02-10
Stream fixes

gasche:
Fixing `stream_to_file` bugs reported by @tcoopman in https://github.com/ocaml-community/yojson/pull/131#issuecomment-1034114373. The first commit is a clear bugfix (without it, the function is completely broken; I think it has been since 931050073020064a2182e659fd8d09f8a7e8240f), the second is rather an improvement in behavior -- the memory usage of the `stream_to_{file,channel}` functions now corresponds to the size of the largest element in the stream, instead of the total size of all elements of the stream.

### DIFF
--- a/lib/write.ml
+++ b/lib/write.ml
@@ -489,8 +489,11 @@ let stream_to_channel ?buf ?(len=2096) ?std oc st =
         None -> Buffer.create len
       | Some ob -> ob
   in
-  stream_to_buffer ?std ob st;
-  Buffer.output_buffer oc ob
+  Stream.iter (fun json ->
+    to_buffer ?std ob json;
+    Buffer.output_buffer oc ob;
+    Buffer.clear ob;
+  ) st
 
 let stream_to_file ?len ?std file st =
   let oc = open_out file in
--- a/lib/write.ml
+++ b/lib/write.ml
@@ -489,7 +489,8 @@ let stream_to_channel ?buf ?(len=2096) ?std oc st =
         None -> Buffer.create len
       | Some ob -> ob
   in
-  stream_to_buffer ?std ob st
+  stream_to_buffer ?std ob st;
+  Buffer.output_buffer oc ob
 
 let stream_to_file ?len ?std file st =
   let oc = open_out file in
--- a/test/test_write.ml
+++ b/test/test_write.ml
@@ -29,7 +29,24 @@ let to_file () =
   test ~newline:true ();
   test ~newline:false ()
 
+let stream_to_file () =
+  let output_file = Filename.temp_file "test_yojson_stream_to_file" ".json" in
+  let data = [`String "foo"; `String "bar"] in
+  Yojson.Safe.stream_to_file output_file (Stream.of_list data);
+  let read_data =
+    let stream = Yojson.Safe.stream_from_file output_file in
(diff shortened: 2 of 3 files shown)


tcoopman:
One thing I noticed, is that with `js_of_ocaml` if the output channel is not flushed regularly, the memory consumption also keeps growing - even with the fix in this PR.

As flushing the output_channel seems very costly doing it on every iteration is definitely not ok. I'm not sure if this should be part of yojson or just documented (or if it's a bug in js_of_ocaml), but if you're using this code with `js_of_ocaml` and don't want to grow your memory consumption, the fix seems to be to `flush oc` every XX iterations.
See: https://discuss.ocaml.org/t/memory-keeps-growing-with-buffer-and-js-of-ocaml/9299

If anyone knows a better way of handling this, I'd like to know :-)

gasche:
I think the fix for this issue doesn't belong to the `Yojson` layer -- we could flush regularly on the producer side, but this is a change of behavior that is not justified in general. It may be possible to workaround the js_of_ocaml issue on the client side, by doing something like `let seq = Seq.mapi (fun i v -> if i mod n = 0 then flush oc; v) in Yojson.seq_to_channel seq oc`. You probably want a fix on the js_of_ocaml side, though.

lib/write.ml
```diff
@@ -489,8 +489,11 @@ let stream_to_channel ?buf ?(len=2096) ?std oc st =
         None -> Buffer.create len
       | Some ob -> ob
   in
-  stream_to_buffer ?std ob st;
-  Buffer.output_buffer oc ob
+  Stream.iter (fun json ->
+    to_buffer ?std ob json;
+    Buffer.output_buffer oc ob;
+    Buffer.clear ob;
```
Leonidas-from-XIV, line 495: I believe this semicolon is superfluous.

test/test_write.ml
```diff
@@ -29,7 +29,24 @@ let to_file () =
...
+  Yojson.Safe.stream_to_file output_file (Stream.of_list data);
+  let read_data =
+    let stream = Yojson.Safe.stream_from_file output_file in
+    let acc = ref [] in
+    Stream.iter (fun v -> acc := v :: !acc) stream;
+    List.rev !acc
+  in
+  Sys.remove output_file;
+  if data <> read_data then
+    (* TODO: it would be nice to use Alcotest.check,
```
Leonidas-from-XIV, line 44: If you want you can add an instance for `Yojson.Safe.t` if you want, it would be a nice improvement and should be quite simple to do (and actually sensible in the long run, I think the reason why there isn't one is because the tests are mostly translated from OUnit. But this is up for you, a nice to have but not necessary.

Leonidas-from-XIV (approved):
Thanks, these are some really nice changes and I really appreciate that you added a test to make sure it doesn't break in the future anymore.

I was thinking whether there's a good way to test that the memory use doesn't balloon out of proportion but that's quite tricky as the test would need to monitor the sizes of the buffers or something like this, which feels rather fragile.

gasche:
> Can you add an entry to the Changelog?

Done!

gasche:
Thanks for the merge! I'll work on rebasing #131 next.

PR #134, opened 2022-02-16
Add benchmark for running `to_channel` with a preallocated buffer

Leonidas-from-XIV:
Since I was curious myself how much of a difference it makes to provide a buffer instead of letting Yojson create one (cc @c-cube).

The results on current `master` are as follows:

```
Estimated testing time 20s (2 benchmarks x 10s). Change using '-quota'.
┌───────────────────────────────────┬──────────┬─────────┬────────────┬──────────┬────────────┐
│ Name                              │ Time/Run │ mWd/Run │   mjWd/Run │ Prom/Run │ Percentage │
├───────────────────────────────────┼──────────┼─────────┼────────────┼──────────┼────────────┤
│ JSON writing with internal buffer │ 594.23us │ 20.02kw │ 32_269.74w │    1.74w │    100.00% │
│ JSON writing with provided buffer │ 498.96us │ 20.02kw │     -3.90w │    0.70w │     83.97% │
└───────────────────────────────────┴──────────┴─────────┴────────────┴──────────┴────────────┘
```

While touching the code, `Core_bench` supports comparing two different implementations so I split the benchmark apart into multiple subcommands. It is probably sensible run different test-runs with a "fresh" runtime instead of running everything in the same runtime, hence each run starts the test binary from scratch.

### DIFF
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -32,8 +32,8 @@ let streamable_string =
   done;
   Buffer.contents buf
 
-let main () =
-  Command.run (Bench.make_command [
+let generic =
+  Bench.make_command [
     Bench.Test.create ~name:"JSON reading" (fun () ->
       ignore (Yojson.Safe.from_string data));
     Bench.Test.create ~name:"JSON writing" (fun () ->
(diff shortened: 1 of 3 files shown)


bench/bench.ml
```diff
@@ -60,7 +60,28 @@ let main () =
...
+  let buf = Buffer.create 4096 in
+  let data = large_int_assoc in
+  Bench.make_command [
+    Bench.Test.create ~name:"JSON writing with internal buffer" (fun () ->
+      Out_channel.with_file "/dev/null" ~f:(fun oc ->
+        ignore (Yojson.Safe.to_channel oc data)));
+    Bench.Test.create ~name:"JSON writing with provided buffer" (fun () ->
+      Out_channel.with_file "/dev/null" ~f:(fun oc ->
+        (* make extra sure the buffer is empty *)
+        Buffer.clear buf;
```
panglesd, line 75: Maybe, when #132 is merged, these two lines can be removed?
As it's a benchmark, we do not want to do extra steps than what is necessary.
I think we should wait for #132 to be merged, and rebase this PR on top of it, without these lines.

panglesd (approved):
Looks good to me!
A good thing that you added the framework for adding more benchmarks.
As I said in the comment, maybe we should rebase this on top of #132 when it is merged, and avoid duplicating the buffer cleaning.

bench/bench.ml
Leonidas-from-XIV: I merged #132 and rebased upon it, thus removing the `Buffer.clear`.

PR #136, opened 2022-03-04
Add validate_t

metanivek:
While experimenting with atdgen, I came across an edge case when trying validations and untypable JSON.

If I use the following, I get a build warning about deprecation of the `json` type.
```
type json <ocaml module="Yojson.Safe"> = abstract
```

If I add an explicit type (`t="t"`) to address the warning, compilation fails with `Unbound value Yojson.Safe.validate_t`, so this PR adds it. Since it appears that the `json` type is disappearing, I also made `validate_json` an alias to `validate_t` as the canonical definition.

### DIFF
--- a/lib/read.mli
+++ b/lib/read.mli
@@ -261,11 +261,13 @@ val read_json : lexer_state -> Lexing.lexbuf -> t
 val skip_json : lexer_state -> Lexing.lexbuf -> unit
 val buffer_json : lexer_state -> Lexing.lexbuf -> unit
 
+val validate_t : 'path -> t -> 'error option
 val validate_json : 'path -> t -> 'error option
   (* always returns [None].
      Provided so that atdgen users can write:
 
        type t <ocaml module="Yojson.Safe"> = abstract
+       type json <ocaml module="Yojson.Safe"> = abstract
   *)
 
 (* end undocumented section *)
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV:
The type `json` will be removed removed in 2.0.0, atdgen will need to update to use `Yojson.Safe.t`.

In fact, I feel like we should probably also remove `validate_json` since this function doesn't seem to serve any purpose, @panglesd?

metanivek:
Got it. I’m happy to remove `validate_json` here. atdgen does work with `Yojson.Safe.t` but it needs `validate_t` to exist when using validations.

Leonidas-from-XIV:
I think in such case it would be more reasonable to define a module that wraps Yojson.Safe.t in your code and provides `validate_t`. Or change atdgen to allow injecting a validation function in another way.

metanivek:
Agreed. This seems like a good opportunity to break this connection between the two libraries since it is really an atdgen concern.

MisterDA:
Just had the same problem. Could you explain how to turn off the validation for the json type? Every combination I'm using turn out to be wrong with the code atdgen.2.2.1 generates.
You could also point to this issue from the changelog.

metanivek:
@MisterDA I missed your comment previously, but I came back to this the other day and realized it wasn't working. I made a PR that should fix this as of [atdgen.2.9.1 release](https://github.com/ahrefs/atd/releases/tag/2.9.1). 🎉

PR #137, opened 2022-03-07
Remove `validate_json`

Leonidas-from-XIV:
It doesn't do much and made only sense in atdgen when we used to have a `json` type. Now that the type doesn't exist it would require to be renamed into `validate_t` but such functionality doesn't seem to make sense to exist in Yojson, especially as it never does any validation to begin with.

This issue was brought up by @metanivek in #136 and my assumption is that if we're changing the API then this function has no reason to exist and validation should be done on the atdgen side.


panglesd (approved):
LGTM. Seems a good idea to remove old atdgen specific code.
Thanks @metanivek and @Leonidas-from-XIV !

mjambon:
Link to relevant ATD issue: https://github.com/ahrefs/atd/issues/263

PR #138, opened 2022-03-23
Use `Command_unix` following the deprecation of `Command`

Leonidas-from-XIV:
In Core v0.15 `Command.run` is deprecated in favor of `Command_unix` to be able to split Unix parts into their separate packages. This means the code needs to adapt. Turns out v0.14 supports that already and since this is a benchmarking binary there isn't that much value in keeping it installable on large range of versions of Core.

### DIFF
--- a/yojson-bench.opam
+++ b/yojson-bench.opam
@@ -7,11 +7,12 @@ dev-repo: "git+https://github.com/ocaml-community/yojson.git"
 doc: "https://ocaml-community.github.io/yojson/"
 license: "BSD-3-Clause"
 depends: [
-  "ocaml" {>= "4.04"}
+  "ocaml" {>= "4.08"}
   "yojson" {= version}
   "dune"
-  "core_bench" {>= "v0.11.0"}
-  "core" {>= "v0.11.0"}
+  "core_bench" {>= "v0.14.0"}
+  "core" {>= "v0.14.0"}
+  "core_unix" {>= "v0.14.0"}
 ]
 build: [
   ["dune" "subst"] {pinned}
--- a/bench/bench.ml
+++ b/bench/bench.ml
@@ -79,7 +79,7 @@ let main () =
     ("generic", generic);
     ("buffer", buffer)
   ]
-  |> Command.run
+  |> Command_unix.run
 
 let () =
   main ()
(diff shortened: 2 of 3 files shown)


yojson-bench.opam
```diff
@@ -7,11 +7,12 @@ dev-repo: "git+https://github.com/ocaml-community/yojson.git"
 doc: "https://ocaml-community.github.io/yojson/"
 license: "BSD-3-Clause"
 depends: [
-  "ocaml" {>= "4.04"}
+  "ocaml" {>= "4.08"}
```
panglesd, line 10: Why adding this constraint? It is not inherent to `yojson-bench`, and I don't think it should be here (eg if core `0.16` supports `ocaml.4.04` for instance!). `opam` will generate this dependency by transitivity anyway.

panglesd (commented):
I approve except for the one small comment I made, which I hope makes sense!

yojson-bench.opam
Leonidas-from-XIV: This is because `core_bench` [supports 4.08 at minimum](https://opam.ocaml.org/packages/core_bench/core_bench.v0.14.0/).

My reasoning here is that if it can't be installed on 4.04 (because even although the code compiles, the dependencies don't) then we can't really test it on such old releases (e.g. via the minimal version mode of OCaml CI) and thus can't make sure our code actually builds on 4.04, the lowest we can check is 4.08.

c-cube:
But core_bench isn't a real dependency, is it? A solution for CI is to have one that covers many versions with just build + test, and then one that runs benchmarks on a recent ocaml. I think 4.08 is fine but it's sad to drop 4 versions just for a dev dependency :-)

c-cube:
My bad, that's what happens when I reply from my phone :sweat_smile: . Apologies for the disruption.

panglesd (approved):

yojson-bench.opam
panglesd: Thanks for the explanation! I would have said that it's not a problem to not be able to test, as anyway it cannot be installed. If at some point `core_bench` required `ocaml >= 4.04`, we would be automatically warned by the CI if it works or not.

But that is very minor thing and I rally easily to your point of view!

PR #139, opened 2022-06-02
Fix formatting bug in documentation

bcc32:
I noticed that the heading for `safe` wasn't appearing correctly on the generated HTML, see https://ocaml-community.github.io/yojson/yojson/Yojson/index.html#basic.

### DIFF
--- a/lib/yojson.cppo.mli
+++ b/lib/yojson.cppo.mli
@@ -44,7 +44,7 @@ end
 #undef STRING
 end
 
-(** {{1:safe} Multipurpose JSON tree type} *)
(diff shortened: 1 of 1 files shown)


panglesd:
Thanks @bcc32 !
This seems to have been already fixed in 9f53260133e85a42beac109dad24015e8e8e5ec4

bcc32:
Thanks, closing.

Leonidas-from-XIV:
Thanks for the contribution. It should be part of the 2.0.1 release which I assume is gonna happen soon since a lot of small issues are piling up.

PR #140, opened 2022-06-03
Add regression test for pretty-printing

Leonidas-from-XIV:
This adds the JSON that ATD uses and checks whether the result looks as expected.

Of course the details are in the eye of the beholder but the reason this is added is because Yojson 2.0 pretty-prints pretty ugly structures and we hope to improve on that.


c-cube:
nice! it's useful to see where nested objects print badly.

c-cube (approved):

PR #141, opened 2022-06-03
improve pretty printing

c-cube:
cf #140

PR #142, opened 2022-06-07
Wrap arrays using an hov box only if they contain only simple elements.

mjambon:
This produces a similar output to yojson < 2. See the expected test output.

This is a follow-up to https://github.com/ocaml-community/yojson/pull/141#issuecomment-1148007694


test/pretty/atd.expected.json
```diff
@@ -1,12 +1,35 @@
...
   [
-    "V1", "v22", [ "V3", "testing" ], [ "V44", 255 ], [ "V5", "None" ],
-    [ "V5", [ "Some", true ] ] ], { "v2": "A" }, { "v2": [ "B", 100 ] },
-  [ "C1", [ "C2", true ], [ "C2", false ] ], [ 50, 30, -1, 400 ]
+    "V1",
+    "v22",
+    [ "V3", "testing" ],
+    [ "V44", 255 ],
+    [ "V5", "None" ],
+    [ "V5", [ "Some", true ] ] ],
```
Leonidas-from-XIV, line 30: Shouldn't the last closing `]` here be on its own line?

Leonidas-from-XIV (commented):
Looks great and definitely an improvement, thanks for contributing!

While looking through the expected value, I ran into an odd case, where I think the code formats incorrectly, the trailing `]` of a non-atom that's inside a box seems misplaced.

test/pretty/atd.expected.json
c-cube: I agree, it's surprising.

mjambon: @c-cube I'm trying to figure out what `inside_box` is for. From looking that the easy-format library, it seems that what we want to print JSON arrays is:
* an outer hv box (horizontal or vertical layout) rather than hov (current code when `not inside_box`).
* an inner box for the list body, that gets indented (hv or hov depending on the chosen style).

I'll try a few things.

mjambon: Changing the array's outer hov box to hv seems to have worked.

Leonidas-from-XIV:
Excellent, thanks for the fix! It makes the output look much better.

PR #143, opened 2022-06-09
`dune subst` should be only called when `dev`

Leonidas-from-XIV:
This is according to the opam-repository linter.


panglesd (approved):
If the linter says so!

PR #148, opened 2022-07-17
Expand documentation of exceptions

sim642:
Primarily makes it explicit that parsing errors raise `Json_error`, which isn't on the same page as parsing functions and didn't have any description. Surprisingly, serialization may also raise it in a corner case.
Secondarily, I reviewed all other raised exceptions to consistently document those as well (using `@raise`).

### DIFF
--- a/lib/common.mli
+++ b/lib/common.mli
@@ -1,8 +1,10 @@
 val version : string
 
 exception Json_error of string
+(** Exception (usually) describing a parsing failure. *)
 
 val json_error : string -> 'a
+(** @raise Json_error *)
 
 type lexer_state = {
   buf : Buffer.t;
--- a/lib/common.mli
+++ b/lib/common.mli
@@ -1,7 +1,9 @@
 val version : string
 
 exception Json_error of string
-(** Exception (usually) describing a parsing failure. *)
+(** Exception used:
+    - in JSON readers, if parsing fails;
+    - in JSON writers and pretty printing, if [float] value is not allowed in standard JSON. *)
 
 val json_error : string -> 'a
 (** @raise Json_error *)
--- a/lib/read.mli
+++ b/lib/read.mli
@@ -136,7 +136,7 @@ val lineseq_from_channel :
 
       See [seq_from_channel] for the meaning of the optional [fin]
       argument.
-      See [from_string] for the meaning of the other optional arguments and returned exceptions. *)
+      See [from_string] for the meaning of the other optional arguments and raised exceptions. *)
 
 val lineseq_from_file :
   ?buf:Buffer.t ->
@@ -149,7 +149,7 @@ val lineseq_from_file :
 
       See [seq_from_channel] for the meaning of the optional [fin]
       argument.
-      See [from_string] for the meaning of the other optional arguments and returned exceptions. *)
+      See [from_string] for the meaning of the other optional arguments and raised exceptions. *)
 
 val read_t : lexer_state -> Lexing.lexbuf -> t
 (** Read a JSON value from the given lexer_state and lexing buffer and return it.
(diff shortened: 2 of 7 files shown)


lib/common.mli
```diff
@@ -1,8 +1,10 @@
 val version : string
 
 exception Json_error of string
+(** Exception (usually) describing a parsing failure. *)
```
Leonidas-from-XIV, line 4: I think the "usually" should be qualified a bit better because I think now it is adding more questions than answering those.

lib/read.mli
```diff
@@ -135,7 +136,7 @@ val lineseq_from_channel :
 
       See [seq_from_channel] for the meaning of the optional [fin]
       argument.
-      See [from_string] for the meaning of the other optional arguments. *)
+      See [from_string] for the meaning of the other optional arguments and returned exceptions. *)
```
Leonidas-from-XIV, line 139: "raised exceptions" to be consistent.

lib/read.mli
```diff
@@ -148,7 +149,7 @@ val lineseq_from_file :
...
```
Leonidas-from-XIV, line 152: Same as above.

lib/write.ml
```diff
@@ -121,7 +121,7 @@ let float_needs_period s =
 
   The _fast version is faster but often produces unnecessarily long numbers.
 *)
-let write_float_fast ob x =
+let write_float_fast ob x = (* unused *)
```
Leonidas-from-XIV, line 124: Can't really say that it is unused, since this is in the `.mli`, so it might be used by some code on OPAM.

Though it might very well be that nobody is using it, in which case we should consider removing it.

lib/write.ml
```diff
@@ -175,7 +175,7 @@ let write_normal_float_prec significant_figures ob x =
   if float_needs_period s then
     Buffer.add_string ob ".0"
 
-let write_float_prec significant_figures ob x =
+let write_float_prec significant_figures ob x = (* used by atdgen *)
```
Leonidas-from-XIV, line 178: Please add these comments on the line above, inline comments are nearly always a pain to deal with.

Leonidas-from-XIV (changes requested):
These are some very nice improvements. Thanks for your work investigating and documenting the exceptions - it is rather thankless to figure it out and having an exception raised without being aware of it is definitely a source of frustration.

I pointed out a few notes on how to improve the PR, mostly for consistency/clarity.

lib/common.mli
sim642: I just expanded the description to list both usages.

lib/read.mli
sim642: I initially used "returned" because `lineseq_from_*` doesn't actually raise them but returns with `Exn` instead. I changed it though, because it should be clear from above that all exceptions are via `Exn`.

lib/write.ml
sim642: At least according to [sherlocode](https://sherlocode.com/?q=write_float_fast), nothing on opam uses this. Since it's hidden from the documentation as well, I suppose there doesn't have to be any stability guarantees for this part of the interface. Maybe something at some point needed this (also atdgen?).

If you'd like, I can remove this comment from this PR, but the actual removal of the function is best to be left as a separate matter.

Leonidas-from-XIV: Thanks for the research. I agree removing this would not make sense in the scope of this PR, but if you happen to have a moment after this is merged to remove the unused functions that would be nice. Otherwise I'll probably get to it… eventually.

Of course, there might be some non-public code that uses these functions, but accumulating tons of functions without a clear usecase is somewhat of an anti-pattern and makes maintenance harder.

sim642:
> Can you please add a note to the changelog about your changes (short note, PR number and your GH handle)?

Done!

Leonidas-from-XIV:
Excellent, thanks for the work!

PR #153, opened 2022-08-15
Enable `ocamlformat` on the part of the codebase that doesn't use CPPO

Leonidas-from-XIV:
This enables `ocamlformat` on the subset of the codebase that does not use CPPO.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Introduced ocamlformat
+3101ec295fbc791a1e9384f50178846198a607c7
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,2 @@
+version=0.24.1
+profile=conventional
--- a/.ocamlformat-ignore
+++ b/.ocamlformat-ignore
@@ -1,0 +1,6 @@
+# disable ocamlformat on files that are processed by CPPO
+lib/*.cppo.ml*
+lib/monomorphic.ml
+lib/write.ml*
+lib/pretty.ml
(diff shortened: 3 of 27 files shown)


panglesd (approved):
I approve. I would also suggest adding a `.git-blame-ignore-revs` so that `git blame` output more meaningful result.

panglesd (changes requested):
In fact, currently not only `.cppo.ml` contains cppo instructions, see for instance `monomorphic.ml`.

They should either be renamed, or added to the `ocamlformat-ignore`!

Leonidas-from-XIV:
Oh yes, that's true, you're right. I somehow assumed only the `.cppo.` files had CPPO in them, so I added them to the ignore-file to avoid churn. Turns out `ocamlformat` didn't reformat them anyway (so I didn't have to undo anything), maybe it has gotten sufficiently smart to skip them. Still good to have it skip over.

Also, I added the revision of the commit to the file. Unfortunately, there's no default name that git would pick up and users need to configure git for this to work but adding it won't hurt.

panglesd (approved):
Looks good!

> Also, I added the revision of the commit to the file. Unfortunately, there's no default name that git would pick up and users need to configure git for this to work but adding it won't hurt.

I think `.git-blame-ignore-revs` is the name that github picks, so it's at least directly useful in the github interface. (https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

Leonidas-from-XIV:
Ah, that's great to know, definitely useful!

PR #154, opened 2022-08-16
Generate OPAM files from dune-project

Leonidas-from-XIV:
This has some advantages, e.g. auto-generating the correct dune incantation as well as deduplicating some of the OPAM stanzas.


panglesd (approved):
I think it is an improvement!

PR #157, opened 2022-10-28
Add path function

cuihtlauac:
Allows recurisve access to nested object fields using a list of keys

### DIFF
--- a/lib/util.mli
+++ b/lib/util.mli
@@ -85,12 +85,9 @@ val member : string -> t -> t
       object [obj], or [`Null] if [k] is not present in [obj].
       @raise Type_error if [obj] is not a JSON object. *)
 
-val path : string list -> t -> t
-(* [path l obj] returns the value associated with the list of keys [l] in
-     the JSON object [obj], or [`Null] if [l] did no led to an object in
-     [obj]. [path [k1; ...; kn] obj] is the same as
-     [member kn (... (member k1 obj))]
-     @raise Type_error if [obj] is not a JSON object. *)
+val path : string list -> t -> t option
+(* [path l obj] recurses the JSON object [obj] for each key in the path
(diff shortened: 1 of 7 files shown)


lib/util.mli
```diff
@@ -85,6 +85,13 @@ val member : string -> t -> t
       object [obj], or [`Null] if [k] is not present in [obj].
       @raise Type_error if [obj] is not a JSON object. *)
 
+val path : string list -> t -> t
+(* [path l obj] returns the value associated with the list of keys [l] in
+     the JSON object [obj], or [`Null] if [l] did no led to an object in
+     [obj]. [path [k1; ...; kn] obj] is the same as
+     [member kn (... (member k1 obj))]
+     @raise Type_error if [obj] is not a JSON object. *)
```
Leonidas-from-XIV, line 93: I think in 2022 we shouldn't be adding more unsafe functions that throw exceptions. It would be better to return `t option` here and I am somewhat unsure that returning `` `Null`` is a behavior that makes sense in OCaml. If I saw a function like this I would not expect `path ["missing"] (`Assoc [])` to return `` `Null``, I would rather expect it to throw `Not_found` following the stdlib.

(I am aware that it uses `member` for which the exact same concerns apply which unfairly gets a free pass for being like this since forever)

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
@panglesd Can you take a look at this PR and weight in whether you think this is reasonable? It's a bit unusual as the API is bit more exception-safe than the rest.

panglesd (approved):
We are now in 2023 and I think we should even less add unsafe API than in 2022.
Consistency with the rest of the API seems, to me, less important than having safer functions.

The `string list -> t -> t option` and its implementation look good to me.

PR #158, opened 2022-12-27
Remove Tuple & Variant

### LINKED ISSUE #104: [RFC] Remove support for `Tuple` and `Variant`
# Current state

Yojson currently supports proprietary extensions to JSON, namely using `(x, y)` syntax to represent tuples and `<Bar:"abc">` for variants.

# Proposal & Rationale

JSON is mainly a language-neutral format for information exchange, not much for serialization of data and as such it makes sense to support a language-interoperable set of JSON grammar. As such, outputting data using these features will yield problems with other parsers that do not understand that grammar extension.

As such the suggestion is to remove these variants from the `Yojson.{Basic,Safe,Raw}.t` type, remove the pretty printing support and remove the parsing code.

Removing the code makes the maintenance easier, is simpler for people to understand (people knowing JSON might be confused why `` `Tuple`` or `` `Variant`` can be produced) and makes it easier to work on the parser since it can match the grammar of RFC 8259 without fitting in proprietary data types.

# Impact

While this is a breaking change, we estimate the impact to be low, since presumably most JSON parsed with Yojson comes from non-OCaml sources that do not use these non-standard extensions.

The most obvious places that might break in the OCamlverse might be the schema generation tools:

* `ppx_deriving_yojson` does not use `` `Tuple`` nor `` `Variant``, it encodes both tuples and variants as `` `List``s
* `ppx_yojson_conv` does not seem to use `` `Tuple`` nor `` `Variant``
* `atd` seems to encode things using `` `Variant`` ([source](https://github.com/ahrefs/atd/blob/574fd482bf9ae82df21f8e32d4929f1a99d35cac/atdgen-codec-runtime/src/json.ml)), but it might be possible to switch to an encoding that's more like what the `deriving`-plugins do? Is there someone who we could talk on atd's behalf? @rgrinberg maybe?

# Roadmap

This would be planned for a 3.0.0 release at the earliest, to give people enough time to discuss the impact and update their code. We could possibly try a rebuild of the revdeps in the duniverse to see how many projects are affected.

Leonidas-from-XIV:
It has been long time on my mind but now that Yojson 2.x is finally out, we can start looking at the next step, Yojson 3.0 (which I assume should be a fairly minor issue issue, given most consumers probably don't ever deal with `` `Tuple`` or `` `Variant``).

The code itself is pretty self-explanatory, it mostly eliminates the `TUPLE` and `VARIANT` CPPO variables and then deletes some support code.

There might be more code to be deleted, like `start_any_variant`/`start_any_tuple`. I assume these might be used by atd (need to look into it) so I just removed the part that parses non-standard syntax but maybe it can be removed altogether, along with the unit test that covers it.

Closes #104

### DIFF
--- a/lib/read.mll
+++ b/lib/read.mll
@@ -771,15 +771,6 @@ and finish_buffer_stringlit v = parse
   | _    { long_error "Invalid string literal" v lexbuf }
   | eof  { custom_error "Unexpected end of input" v lexbuf }
 
-and finish_buffer_variant v = parse
-    ':'  { Buffer.add_char v.buf ':';
-           buffer_json v lexbuf;
-           buffer_space v lexbuf;
-           buffer_gt v lexbuf }
-  | '>'  { Buffer.add_char v.buf '>' }
-  | _    { long_error "Expected ':' or '>' but found" v lexbuf }
-  | eof  { custom_error "Unexpected end of input" v lexbuf }
-
 and buffer_ident v = parse
     '"'      { finish_buffer_stringlit v lexbuf }
   | ident    { add_lexeme v.buf lexbuf }
--- a/lib/read.mll
+++ b/lib/read.mll
@@ -250,14 +250,6 @@ rule read_json v = parse
                      `List (List.rev !acc)
                  }
 
-  | '('          {
-                   long_error "Invalid token" v lexbuf
-                 }
-
-  | '<'          {
-                   long_error "Invalid token" v lexbuf
-                 }
-
   | "//"[^'\n']* { read_json v lexbuf }
   | "/*"         { finish_comment v lexbuf; read_json v lexbuf }
   | "\n"         { newline v lexbuf; read_json v lexbuf }
@@ -683,14 +675,6 @@ and skip_json v = parse
                      ()
                  }
 
-  | '('          {
-                   long_error "Invalid token" v lexbuf
-                 }
-
-  | '<'          {
-                   long_error "Invalid token" v lexbuf
-                 }
-
   | "//"[^'\n']* { skip_json v lexbuf }
   | "/*"         { finish_comment v lexbuf; skip_json v lexbuf }
   | "\n"         { newline v lexbuf; skip_json v lexbuf }
@@ -764,14 +748,6 @@ and buffer_json v = parse
                      assert false
                    with Common.End_of_array ->
                      ()
-                 }
-
-  | '('          {
-                   long_error "Invalid token" v lexbuf
-                 }
-
-  | '<'          {
-                   long_error "Invalid token" v lexbuf
                  }
 
   | "//"[^'\n']* { add_lexeme v.buf lexbuf; buffer_json v lexbuf }
--- a/lib/read.mll
+++ b/lib/read.mll
@@ -146,8 +146,6 @@
   let map_lexeme f (lexbuf : Lexing.lexbuf) =
     let len = lexbuf.lex_curr_pos - lexbuf.lex_start_pos in
     f (Bytes.sub_string lexbuf.lex_buffer lexbuf.lex_start_pos len) 0 len
-
-  type variant_kind = [ `Square_bracket | `Double_quote ]
 }
 
 let space = [' ' '\t' '\r']+
@@ -338,13 +336,6 @@ and read_comma v = parse
     ','  { () }
   | _    { long_error "Expected ',' but found" v lexbuf }
   | eof  { custom_error "Unexpected end of input" v lexbuf }
-
-and start_any_variant v = parse
-  | '"'      { Buffer.clear v.buf;
-               `Double_quote }
-  | '['      { `Square_bracket }
-  | _        { long_error "Expected '<', '\"' or '[' but found" v lexbuf }
-  | eof      { custom_error "Unexpected end of input" v lexbuf }
 
 and finish_comment v = parse
   | "*/" { () }
@@ -596,11 +587,6 @@ and read_object_sep v = parse
 and read_colon v = parse
     ':'      { () }
   | _        { long_error "Expected ':' but found" v lexbuf }
-  | eof      { custom_error "Unexpected end of input" v lexbuf }
-
-and start_any_tuple v = parse
-  | '['      { true }
-  | _        { long_error "Expected '(' or '[' but found" v lexbuf }
   | eof      { custom_error "Unexpected end of input" v lexbuf }
 
 and read_lpar v = parse
--- a/lib/read.mll
+++ b/lib/read.mll
@@ -147,7 +147,7 @@
     let len = lexbuf.lex_curr_pos - lexbuf.lex_start_pos in
     f (Bytes.sub_string lexbuf.lex_buffer lexbuf.lex_start_pos len) 0 len
 
-  type variant_kind = [ `Edgy_bracket | `Square_bracket | `Double_quote ]
+  type variant_kind = [ `Square_bracket | `Double_quote ]
 }
 
 let space = [' ' '\t' '\r']+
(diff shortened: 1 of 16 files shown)


lib/read.mll
```diff
@@ -268,35 +268,11 @@ rule read_json v = parse
                  }
 
   | '('          {
```
hhugo, line 270: This case can be removed

lib/read.mll
```diff
@@ -268,35 +268,11 @@ rule read_json v = parse
...
-                       assert false
-                     with End_of_tuple ->
-                       `Tuple (List.rev !acc)
-                   #else
-                     long_error "Invalid token" v lexbuf
-                   #endif
+                   long_error "Invalid token" v lexbuf
                  }
 
   | '<'          {
```
hhugo, line 274: This case can be removed

lib/read.mll
```diff
@@ -797,34 +700,11 @@ and skip_json v = parse
...
-                       assert false
-                     with End_of_tuple ->
-                       ()
-                   #else
-                     long_error "Invalid token" v lexbuf
-                   #endif
+                   long_error "Invalid token" v lexbuf
                  }
 
   | '<'          {
```
hhugo, line 706: can be removed

lib/read.mll
```diff
@@ -911,36 +783,11 @@ and buffer_json v = parse
                  }
 
   | '('          {
```
hhugo, line 785: can be removed

lib/read.mll
```diff
@@ -911,36 +783,11 @@ and buffer_json v = parse
...
-                       assert false
-                     with End_of_tuple ->
-                       ()
-                   #else
-                     long_error "Invalid token" v lexbuf
-                   #endif
+                   long_error "Invalid token" v lexbuf
                  }
 
   | '<'          {
```
hhugo, line 789: can be removed

c-cube:
Is this going to kill `ppx_deriving_yojson`  for good? :)

Leonidas-from-XIV:
@c-cube  I don't think it will break ppx_deriving_yojson too bad, as it encodes tuples as lists and variants as lists with the name of the constructor as first argument, if I remember correctly. Here's the code: https://github.com/ocaml-ppx/ppx_deriving_yojson/blob/61f2d63138ec1efea0a3fb4afa682f40497ec380/src/ppx_deriving_yojson.ml#L115-L148

The biggest (only?) issue will probably be Atd which uses Tuple and Variant, but I believe it would make sense in that case to adopt the same encoding strategy that ppx_deriving_yojson uses. Which has the advantage that the JSON it generates is standard-compliant and thus more interoperable with non-OCaml tools.

mjambon:
> The biggest (only?) issue will probably be Atd which uses Tuple and Variant

I don't think it's a big issue, and it's a good move forward.

Atdgen will make its `-j-std` option the default, and the old default behavior that was supporting the experimental syntax for tuples and variants will be removed.

To see the dependencies on Yojson function, I used this:
```
~/atd/atdgen/src $ git grep -oP 'Yojson.[^" ]+' | sort | uniq -c
      2 ocaml.ml:Yojson.Safe.t
      1 oj_emit.ml:Yojson.End_of_object
      5 oj_emit.ml:Yojson.End_of_tuple
      1 oj_emit.ml:Yojson.Safe.finish_string
      1 oj_emit.ml:Yojson.Safe.init_lexer
      2 oj_emit.ml:Yojson.Safe.lexer_state
      1 oj_emit.ml:Yojson.Safe.map_ident
      1 oj_emit.ml:Yojson.Safe.read_comma
      2 oj_emit.ml:Yojson.Safe.read_gt
      1 oj_emit.ml:Yojson.Safe.read_ident
      1 oj_emit.ml:Yojson.Safe.read_lcurl
      2 oj_emit.ml:Yojson.Safe.read_null_if_possible
      1 oj_emit.ml:Yojson.Safe.read_object_end
      1 oj_emit.ml:Yojson.Safe.read_object_sep
      1 oj_emit.ml:Yojson.Safe.read_rbr
     16 oj_emit.ml:Yojson.Safe.read_space
      1 oj_emit.ml:Yojson.Safe.read_tuple_end2
      4 oj_emit.ml:Yojson.Safe.read_tuple_sep2
      2 oj_emit.ml:Yojson.Safe.skip_json
      1 oj_emit.ml:Yojson.Safe.start_any_tuple
      1 oj_emit.ml:Yojson.Safe.start_any_variant
      1 oj_emit.ml:Yojson.Safe.to_string
      1 oj_emit.ml:Yojson.Safe.write_bool
      1 oj_emit.ml:Yojson.Safe.write_float
      1 oj_emit.ml:Yojson.Safe.write_float_prec
      1 oj_emit.ml:Yojson.Safe.write_int
      1 oj_emit.ml:Yojson.Safe.write_json
      1 oj_emit.ml:Yojson.Safe.write_null
      1 oj_emit.ml:Yojson.Safe.write_std_float
      1 oj_emit.ml:Yojson.Safe.write_std_float_prec
      1 oj_emit.ml:Yojson.Safe.write_string
      1 omelange_emit.ml:Yojson.Safe.to_string
```
Occurrences such as `Yojson.End_of_tuple`, `Yojson.Safe.read_tuple_end2`, and a few others will have to be removed.

See https://github.com/ahrefs/atd/issues/412

Leonidas-from-XIV:
@mjambon That's a great list, will be very handy!

I've been looking at the `std` variants and with the removal of `tuple` and `variant` the only non-standard thing seems to be the printing of floats (`NaN`, `+Infinity`, `-Infinity`). I see that `atdgen` is currently emitting both `write_float(_prec)` and `write_std_float(_prec)`. Looking at `atdgen` it seems to be configurable (`p.std`) which variant it will use.

I'm thinking that as we're breaking compatibility anyway to become more JSON-compliant, might be sensible to bite the bullet and make the `std` variants the only option, thus failing to encode unrepresentable floats in JSON. WDYT?

barracuda156:
This has broken some dependents of `yojson`: https://github.com/tmcgilchrist/ocaml-gitlab/issues/101

mjambon:
I'm discovering there's been some miscommunication here. While in atdgen we're moving away from non-standard JSON syntax (using `<` and `>` for variants, and `(` and `)`) for tuples, we still need a function like the old `start_any_variant` that reads either a double quote that starts a string or a square bracket that starts an array.

The generated code was:
```ocaml
    match Yojson.Safe.start_any_variant p lb with
      | `Edgy_bracket -> (
          ...
        )
      | `Double_quote -> (
          match Yojson.Safe.finish_string p lb with
            ...
        )
      | `Square_bracket -> (
          match Atdgen_runtime.Oj_run.read_string p lb with
            ...
        )
```
and we're moving toward:
```ocaml
    match Yojson.Safe.start_any_variant p lb with
      | `Double_quote -> (
          match Yojson.Safe.finish_string p lb with
            ...
        )
      | `Square_bracket -> (
          match Atdgen_runtime.Oj_run.read_string p lb with
            ...
        )
```

I'm looking for solutions. Here are some ideas:
1. Put back `start_any_variant` into Yojson so that it returns a ``[ `Double_quote | `Square_bracket]``.
2. Do nothing to Yojson, and vendor it as part of atdgen = copy it and make minor modifications such as renaming the root module to avoid conflicts (or maybe use a Git submodule).
3. Do both (1) and (2) in parallel, then get rid of (2) once (1) is released.
4. Do nothing to Yojson but fork it and simplify it to make it permanently part of atdgen-runtime.
5. Do nothing to Yojson or Atdgen which will continue to depend on Yojson 2. Implement [atdml](https://github.com/ahrefs/atd/issues/313). This would be a big project (3 full weeks of my time for an initial release with essential features only) that makes a lot of things easier later. One of the features of such a reimplementation is that it would use the normal Yojson parser and translate the AST into specific OCaml data structures. I don't have the budget for this at the moment.

kit-ty-kate:
@mjambon I'm not sure this is much of a problem. The `start_any_variant` function is very easy to reimplement on your side. It's just a .mll file with:
```
rule start_any_variant v = parse
  | '"'      { Buffer.clear v.Yojson__read.buf;
               `Double_quote }
  | '['      { `Square_bracket }
  | _        { [your error case ...] }
  | eof      { [your error case ...] }
```

PR #163, opened 2023-03-31
Raw.Util module CPP

tmcgilchrist:
Fix for https://github.com/ocaml-community/yojson/issues/96

PR #164, opened 2023-04-18
Make `ocamlformat` ignore files using CPPO

Leonidas-from-XIV:
#163 created a CI failure, this should fix it.

(Hope to get rid of CPPO in the long term, but this is how it is for now)


Leonidas-from-XIV:
Merging as this is a pretty much trivially correct change.

PR #166, opened 2023-07-17
Default-ish dune flags

### LINKED ISSUE #165: Consider using -strict-sequence
or even rely on dune default flags (plus adjustments).

dune default's are currently completely ignored because of
```sexp
(flags (-safe-string)))
```

I would suggest to use the following instead
```
(flags (:standard -w -27-32-34))
```
- `safe-string` is the default nowdays. `unsafe-string` was completely removed starting with ocaml 5.0.  
- warnings 27, 32, 34 currently trigger but are harder to fix in the current workflow. 

Dune `:standard` include `-strict-sequence` witch trigger the following error. It's unclear to me which of the implementation or the signature should be fixed. 

```
$ (cd _build/default && /home/hugo/tezos/_opam/bin/ocamlopt.opt -strict-sequence -safe-string -g -I lib/.yojson.objs/byte -I lib/.yojson.objs/native -I /home/hugo/tezos/_opam/lib/seq -intf-suffix .ml -no-alias-deps -opaque -o lib/.yojson.objs/native/yojson.cmx -c -impl lib/yojson.ml)
File "write.ml", line 428, characters 2-62:
Error: This expression has type int but an expression was expected of type
         unit
       because it is in the left-hand side of a sequence
Command [2] exited with code 2:              
$ (cd _build/default && /home/hugo/tezos/_opam/bin/ocamlc.opt -strict-sequence -safe-string -g -bin-annot -I lib/.yojson.objs/byte -I /home/hugo/tezos/_opam/lib/seq -intf-suffix .ml -no-alias-deps -opaque -o lib/.yojson.objs/byte/yojson.cmo -c -impl lib/yojson.ml)
File "write.ml", line 428, characters 2-62:
Error: This expression has type int but an expression was expected of type
         unit
       because it is in the left-hand side of a sequence
```

Leonidas-from-XIV:
Closes #165.


Leonidas-from-XIV:
The CI failure has been reported at https://github.com/ocaml/infrastructure/issues/60. There's still some weirdness about warning 50, since older OCaml version don't like some of our documentation comments, whereas newer versions (and on my local machine) work fine and don't report issues.

Leonidas-from-XIV:
I'll probably rebase this on #167 when that one is merged to look at the failures due to warning 50 on some older OCaml versions in more detail.

PR #168, opened 2023-07-20
Remove forward ref in the lexer

### LINKED ISSUE #84: Large output size on jsoo
When using this library with jsoo, a single reference to the library adds 64k to the js output. This is 20% of the total size of the application. This happens for instance with just a single definition of a function like `let x = Yojson.Basic.to_string`, with no uses. I'm not sure if this is an issue with jsoo that would require a more advanced optimizer, or with dependencies of the library that could perhaps be reduced to avoid pulling in so much code, but since json is a built-in feature of browsers, one would expect it to have relatively low overhead.

hhugo:
The forward ref make jsoo deadcode elimination ineffective.

Fix #84 to some degree.

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -1,4 +1,4 @@
-(ocamllex read)
+(ocamllex read lexer_utils)
 
 (rule
  (targets t.ml)
@@ -111,7 +111,7 @@
 (library
  (name yojson)
  (public_name yojson)
- (modules yojson t basic safe raw common codec)
+ (modules yojson t basic safe raw common codec lexer_utils)
  (synopsis "JSON parsing and printing")
  (libraries seq)
  (flags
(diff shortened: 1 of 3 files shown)


hhugo:
@Leonidas-from-XIV, feel free to take over this PR.

Leonidas-from-XIV (approved):
Looks reasonable to me and honestly a bit nicer than the `optjunk32` business.

I think this area of Yojson is under-tested since we have any tests on parsing errors.

hhugo:
@Leonidas-from-XIV, can this be merged ?

hhugo:
This PR is ready from my perspective

Leonidas-from-XIV:
Thanks for the contribution! Might be worth cutting a patch release to help JSOO users.

PR #169, opened 2023-08-15
accept numbers with leading zeroes (in new module)

ferminr:
Description.

The json spec does not allow leading zeroes in literal numbers, and that's what the yojson parser implements. However, I sometimes consume json data from external sources that does have numbers with leading zeroes and I'd like to be able to parse it. Some json parsers, and some online tools that check correctness of jsonld (eg https://validator.schema.org/) are permissive and accept them too.

Changing the lexer to accept them would be a one-line PR, but a non-backwards compatible change, so I have decided to go for an opt-in approach, with a new module. Here's an example:

```
utop # Yojson.Basic.Util.to_list @@ Yojson.Anynum.Basic.from_string "[01, 00.2]" ;;
- : Yojson.Basic.t list = [`Int 1; `Float 0.2]
```

`Anynum` has submodules `Basic`, `Raw` and `Safe`. The module name `Anynum` is something that can be changed if a better one is suggested.

The PR could include more docs, examples and tests for the new functionality, but I'll wait until a final version of the PR, if it is accepted.

### DIFF
--- a/lib/anynum.cppo.ml
+++ b/lib/anynum.cppo.ml
@@ -1,0 +1,85 @@
+module Basic : sig
+
+#define INT
+#define FLOAT
+#define STRING
+
+#include "type.ml"
+#include "read.mli"
+
+end = struct
+
+#include "type.ml"
+#include "write.ml"
+module Pretty = struct
+  #include "prettyprint.ml"
+end
+
+#include "write2.ml"
+#include "read_anynum.ml"
+
+#undef INT
+#undef FLOAT
+#undef STRING
+end
+
+module Raw : sig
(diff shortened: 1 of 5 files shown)


Leonidas-from-XIV:
Hi! Thanks for your contribution. However I am trying to reduce the amount of parsers (from 4 to preferably 1) and not increase the space of JSON variants, so as long as JSON does not allow leading zeroes, I'd rather not include support for non-standard variants in Yojson.

In addition, numbers starting with `0` are often considered octal literals, so if we added this, users might expect a different behaviour than there is at hand.

Issue #171, opened 2023-10-22
Yojson 2.1.1 errors can have have incorrect locations


johnridesabike:
Example:
```ocaml
let () =
  try ignore @@ Yojson.Basic.from_string {|{"badjson": error}|}
  with Yojson.Json_error s -> print_endline s
```
Output with 2.1.0:
```
Line 1, bytes 12-18:
Invalid token 'error}'
```
Output with 2.1.1:
```
Line 1, bytes 17-18:
Invalid token 'error}'
```

ElectreAAS:
Same problem here (example in utop)
With 2.1.0:
```ocaml
# Yojson.Safe.from_string "hello";;
Exception: Yojson.Json_error "Line 1, bytes 0-5:\nInvalid token 'hello'".
```
With 2.1.1:
```ocaml
# Yojson.Safe.from_string "hello";;
Exception:
Yojson__Common.Json_error "Line 1, bytes 4-5:\nInvalid token 'hello'".
```

Leonidas-from-XIV:
Thanks, that's indeed a regression. Would one of you want to submit a test for it?

The only relevant change I can think of is #168, will need to be looked into.

ElectreAAS:
> Would one of you want to submit a test for it?

On it

Leonidas-from-XIV:
Fixed in #172.

PR #172, opened 2023-11-16
Added failing test on parsing failure

ElectreAAS:
Added a test that currently fails, to highlight the problem raised in #171.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 
 ### Fixed
 
+- Fix the error location reported in the exception. Regression in 2.1.1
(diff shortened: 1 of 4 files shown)


Leonidas-from-XIV (approved):
Thanks, this will come helpful to fix and as regression test!

Leonidas-from-XIV:
It was indeed due to #168, which was advancing the parser as it was reading "junk" (that is the unparseable bits after the parser failure). There is still some wonkiness happening that I would like to look into (adding 1 and subtracting 1 in other places seems a bit odd) but this will at least fix the regression.

PR #175, opened 2024-04-25
Replace CPPO with µCPPO

Leonidas-from-XIV:
This removes a build dependency and replaces the conditional compilation with a mucppo, a CPPO subset meant for embedding in builds: https://github.com/Leonidas-from-XIV/mucppo

The advantage is that it sheds another dependency from Yojson and given it is only needed at compile time, the tool is quickly built (as it is a single file) and never installed into the users switch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@
 
 ### Removed
 
+- Remove CPPO dependency to make the Yojson installation lighter
+  (@Leonidas-from-XIV, #175)
+
 ### Security
 
 ## 2.1.2
--- a/dune-project
+++ b/dune-project
@@ -18,7 +18,6 @@ ydump is a pretty-printing command-line program provided with the
 yojson package.")
   (depends
     (ocaml (>= 4.02.3))
-    (cppo :build)
     (alcotest (and :with-test (>= 0.8.5)))
     (seq (>= 0.2.2))))
 
--- a/lib/dune
+++ b/lib/dune
@@ -1,5 +1,7 @@
 (ocamllex read lexer_utils)
 
+(vendored_dirs mucppo)
+
 (rule
  (targets t.ml)
  (deps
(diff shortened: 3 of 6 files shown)


panglesd:
Just a small comment: If the goal is to remove dependencies, an alternative approach would have been to make the preprocessed output "promoted" in the tree, and directly used when installing through opam. So that (even if we kept using cppo) we could move cppo from the build dependencies to the test dependencies.

We used to have this in odoc: we use ocaml-crunch to generate a module from static files, promoted the result and did not include ocaml-crunch to the dependencies. See the [dune file](https://github.com/ocaml/odoc/blob/5a054ff166b9e30581511baf4caddaa4ffee4c6c/src/html_support_files/dune) at that time. We now more simply include `ocaml-crunch` as a dependency... (https://github.com/ocaml/odoc/pull/1044).

(But your approach is also valid of course!)

Leonidas-from-XIV:
I used such an approach in [xml2js](https://github.com/Leonidas-from-XIV/node-xml2js) to allow people to use the generated source without the CoffeeScript compiler, the issues with this are twofold:

 1. People started making PRs to the generated source, which were entirely unusable as a run of the compiler would overwrite them
 2. In the case of Yojson one needs to run the build anyway to compile the `.ml` sources, so whether the preprocessing happens here makes no difference to the user. In the case of PPX, it would be rather unusual to have the preprocessed source in the tree.

In the long end, my goal would be to get rid of all variants but `Basic` and `Safe`, at which point we might be able to completely get rid of any preprocessing and use e.g. functors or the like. This has the advantage that it works a lot better with Merlin and the code is much easier to read if it doesn't contain flights of code that are not run. For example JSON5 PR is written in a way where it supports both ASTs yet doesn't require any additional conditional compilation.

In the end this is just a small step which removes a (in the grand scheme of things) small dependency, it doesn't preclude any further improvements in the future as you suggest :-)

PR #181, opened 2024-06-03
Use ocaml.org as documentation location

Leonidas-from-XIV:
Instead of writing our own docs to github pages, pointing to ocaml.org is kind of nicer.


Leonidas-from-XIV:
I've talked to the docs team and was encouraged that the online docs are ready for regular use thus merging.

PR #182, opened 2024-06-07
Improve the error messages displayed by the JSON5 lexer and parser.

gcluzel:
Since JSON5 is a standard intended to parse JSON written by humans, it's a bit unfortunate that the error messages were not very descriptive.
I tried to make them closer to the ones displayed by the classic YoJson parser. Also, when it's available, the line where the error message is located is displayed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 *2024-06-04*
 
+### Added
+
+- Add locations in the JSON5 parser error messages.
+
 ### Fixed
 
 - Don't expose `Yojson_five` internals anymore (@Leonidas_from_XIV, #180)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,12 @@
-## 2.2.1
-
-*2024-06-04*
+## unreleased
 
 ### Added
 
-- Add locations in the JSON5 parser error messages.
+- Add locations in the JSON5 parser error messages (@gcluzel, #182)
+
+## 2.2.1
+
+*2024-06-04*
 
 ### Fixed
 
--- a/lib/json5/lexer.ml
+++ b/lib/json5/lexer.ml
@@ -34,15 +34,18 @@ let pp_token ppf = function
   | STRING s -> Format.fprintf ppf "%S" s
   | EOF -> Format.pp_print_string ppf "eof"
 
-let custom_error lexbuf =
+let lexer_error lexbuf =
   let { Lexing.pos_fname = file; pos_lnum = line; _ }, _ =
     Sedlexing.lexing_positions lexbuf
   in
   let file_line =
-    if String.equal file "" then "Line" else Format.sprintf "File %s, line" file
+    if String.equal file "" then "Line" else Printf.sprintf "File %s, line" file
   in
-  Format.sprintf "%s %d: Unexpected character '%s'" file_line line
-    (Sedlexing.Utf8.lexeme lexbuf)
+  let msg =
+    Printf.sprintf "%s %d: Unexpected character '%s'" file_line line
+      (Sedlexing.Utf8.lexeme lexbuf)
+  in
+  Error msg
 
 let source_character = [%sedlex.regexp? any]
 let line_terminator = [%sedlex.regexp? 0x000A | 0x000D | 0x2028 | 0x2029]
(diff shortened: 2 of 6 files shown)


CHANGES.md
```diff
@@ -2,6 +2,10 @@
 
 *2024-06-04*
 
+### Added
+
+- Add locations in the JSON5 parser error messages.
```
Leonidas-from-XIV, line 7: This should include the PR number and your Github handle to credit you.

lib/json5/lexer.ml
```diff
@@ -9,34 +9,40 @@ type token =
...
+  | COLON -> Format.pp_print_string ppf ":"
+  | COMMA -> Format.pp_print_string ppf ","
+  | TRUE -> Format.pp_print_string ppf "true"
+  | FALSE -> Format.pp_print_string ppf "false"
+  | NULL -> Format.pp_print_string ppf "null"
+  | FLOAT s | INT_OR_FLOAT s | INT s -> Format.pp_print_string ppf s
+  | STRING s -> Format.fprintf ppf "%S" s
+  | EOF -> Format.pp_print_string ppf "eof"
+
+let custom_error lexbuf =
```
Leonidas-from-XIV, line 37: I know this is named "custom" in the old parser but a better name would be `lexer_error` and it should presumably return an `Error`

lib/json5/lexer.ml
```diff
@@ -9,34 +9,40 @@ type token =
...
+  let { Lexing.pos_fname = file; pos_lnum = line; _ }, _ =
+    Sedlexing.lexing_positions lexbuf
+  in
+  let file_line =
```
Leonidas-from-XIV, line 41: This logic is duplicated in 3 places, possibly reasonable to join them to be consistent.

lib/json5/lexer.ml
```diff
@@ -9,34 +9,40 @@ type token =
...
+    if String.equal file "" then "Line" else Format.sprintf "File %s, line" file
+  in
+  Format.sprintf "%s %d: Unexpected character '%s'" file_line line
```
Leonidas-from-XIV, line 44: Can use `Printf` instead of `Format`, no `Format` features are used.

lib/json5/parser.ml
```diff
@@ -1,46 +1,71 @@
 open Let_syntax.Result
 
+let custom_error pos error =
```
Leonidas-from-XIV, line 3: `parser_error`?

lib/json5/parser.ml
```diff
@@ -57,12 +82,10 @@ and parse = function
           (Ast.Assoc (List.rev a), xs)
       | x ->
           let s = Format.asprintf "Unexpected token: %a" Lexer.pp_token x in
-          Error s)
+          custom_error pos s)
 
-let parse_from_lexbuf ?fname ?lnum lexbuffer =
-  let fname = Option.value fname ~default:"" in
+let parse_from_lexbuf ?(fname = "") ?(lnum = 1) lexbuffer =
```
Leonidas-from-XIV, line 87: Nice!

Leonidas-from-XIV (changes requested):

Leonidas-from-XIV:
Thanks for the PR, it's definitely useful to have some better error messages. Can you add some tests for the error messages so that we know they are working fine and how they look? That will also serve as regression test.

CHANGES.md
gcluzel: Can you please check the format?

lib/json5/lexer.ml
gcluzel: What do you suggest instead?
I can share the two ones that are in the yojson5 module.

gcluzel:
Thanks for the review.
I have added 4 tests, I let you tell me if you think that it's worth adding other tests. Note that there already are tests that test the parsing error case, they just don't check the error messages displayed in those cases.

CHANGES.md
Leonidas-from-XIV: Looks good now (we have a CI check for it), thanks.

lib/json5/lexer.ml
Leonidas-from-XIV: I've put them in a shared module.

Leonidas-from-XIV:
I think this is good for now. If you want to add more tests you're welcome. The CI failures are known and I have opened https://github.com/ocaml/opam-repository/pull/26154 and https://github.com/ocaml/opam-repository/pull/26156 to solve them but they are unrelated to the PR. Merging.

Thanks for your contribution!

Leonidas-from-XIV (approved):

PR #185, opened 2024-06-28
Remove left-over mentions of tuple and variant

Leonidas-from-XIV:
Turns out in #158 I forgot to update a number of things, mostly issues that the compiler wouldn't warn about. This PR attempts to fix all the remaining bits that can be removed.


Leonidas-from-XIV:
Merging because the failures are all just due to `core_unix` failing to compile on some platforms. Probably needs to be fixed on the OPAM-repository end.

PR #187, opened 2025-03-26
JSON5: fix escape sequences

david-maison-TrustInSoft:
According to the specification https://spec.json5.org/#escapes , as far as I understand : 
- `\'` should be `'` in the returned string (not `\'`)
- `\"` should be `"` (not `\"`)
- and other characters should be returned without the leading backslash too (e.g., `\a` -> `a`)

However, I did not understand what happened for `\1` to `\9`, so I kept the former behavior: rejecting the escape sequence.

### DIFF
--- a/lib/json5/unescape.ml
+++ b/lib/json5/unescape.ml
@@ -66,7 +66,13 @@ let unescape str =
           | None -> Error (Format.sprintf "bad escape sequence %s" escape_chars)
         in
         utf_8_string_of_unicode as_int
-    | 'b' | 'f' | 'n' | 'r' | 't' | 'v' -> Ok str
+    (* https://spec.json5.org/#escapes table 1 *)
+    | 'b' -> Ok "\b"
+    | 'n' -> Ok "\n"
+    | 'r' -> Ok "\r"
+    | 't' -> Ok "\t"
+    | 'f' -> Ok "\x0C"
+    | 'v' -> Ok "\x0B"
     | '0' ->
         if String.length str = 2 then Ok "\x00"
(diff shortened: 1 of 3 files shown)


lib/json5/unescape.ml
```diff
@@ -66,8 +66,7 @@ let unescape str =
           | None -> Error (Format.sprintf "bad escape sequence %s" escape_chars)
         in
         utf_8_string_of_unicode as_int
-    | '"' | '\'' | 'b' | 'f' | 'n' | 'r' | 't' | 'v' -> Ok str
-    | '\\' -> Ok {|\|}
+    | 'b' | 'f' | 'n' | 'r' | 't' | 'v' -> Ok str
```
Leonidas-from-XIV, line 69: I agree that this behavior was wrong, but looking at the table in https://spec.json5.org/#escapes I would expect that we'd need to map `\b` to `U+0008` and not to `\b`?

(It's not a regression but might as well fix it while here?)

Leonidas-from-XIV (commented):
Thanks for your contribution! I agree that this is an issue. I like that there's a test, making sure it is correct but while here I think it would be nice to also properly map the other escape sequences.

lib/json5/unescape.ml
Leonidas-from-XIV: Actually, complaining about this took nearly as long as fixing it, so I pushed a few commits to deal with this.

david-maison-TrustInSoft: Indeed, I don't know why my brain didn't catch that. Sorry, and thanks for the commits!

PR #189, opened 2025-04-25
Replace `core_bench` with `benchmark`

Leonidas-from-XIV:
`core_bench` is nice but the amount of dependency issues and build failures it causes via its `core_unix` dependency is draining a lot of maintenance time trying to get it to work again so this replaces it with a simpler library.

The output is a bit less fancy, but it should allow for similar measurements.


dinosaure:
Sorry to come after your great work but you should take a look on [`bechamel`](https://github.com/mirage/bechamel) which can provides a nice output (with `notty`) and/or an HTML report :+1:. It's actually used for many things for Mirage projects.

Leonidas-from-XIV:
Thanks @dinosaure. @patricoferris pointed me at bechamel before but it looked rather daunting and without a simple CLI option (requiring either `notty` (which itself has a spotty maintenance record) or a browser).

I would be in support of migrating to bechamel at some time in the future, but for now I just wanted to avoid the headache and `benchmark` didn't require any additional dependencies. That also meant that I had to cut out the support for the visualization of the data via `ocurrent-bench` for now, since maintaining it was more trouble than it brought in in benefits.

PR #197, opened 2025-06-12
Adds introductory example

johnwhitington:
Hello!

This pull request adds an introductory example to YoJSON. This is part of a pilot programme funded by the OCaml Software Foundation.

Many OCaml libraries have no examples, or perfunctory examples only. This makes it difficult to get started with a library, particularly if it has an elaborate interface. A working example, no matter how small, can help a newcomer get started quickly. One day, it would be nice to have an example for every Opam package.

For now, examples begin in the OCaml Nursery, here: https://github.com/johnwhitington/ocaml-nursery  (you can read in the README there about the principles behind these examples.) Then, if package authors agree, they are promoted to upstream source. The hope is that this will mean they are more likely to be kept up to date with the library.

The examples are usually, included in a separate directory, within a separate Dune workspace. And so they are intended to be used after installation of the library. However, in this case, I have followed the pattern of the existing examples.

As well as considering accepting this pull request, please do give any comments you have on this programme.

### DIFF
--- a/examples/introduction.json
+++ b/examples/introduction.json
@@ -1,0 +1,3 @@
+{"Number" : 1,
+ "String" : "two",
+ "List": ["containing", "different", "types", 35.4]}
--- a/examples/introduction.ml
+++ b/examples/introduction.ml
@@ -1,13 +1,3 @@
-(* Get a file as a string *)
-let contents_of_file filename =
-  let ch = open_in_bin filename in
-    try
-      let s = really_input_string ch (in_channel_length ch) in
-        close_in ch;
-        s
-    with
-      e -> close_in ch; raise e
-
 (* JSON inline as a string *)
 let json_string =
   {|{"Number" : 1,
@@ -34,7 +24,7 @@ let rec sum_json : Yojson.Safe.t -> float = function
   | _ -> failwith "unexpected construct in sum_json"
   
 let sum filename =
-  Printf.printf "%f\n" (sum_json (Yojson.Safe.from_string (contents_of_file filename)))
+  Printf.printf "%f\n" (sum_json (Yojson.Safe.from_file filename))
 
 (* Read some JSON from a file, reverse any lists in it, and write to file. *)
 let rec reverse_json = function
@@ -43,15 +33,20 @@ let rec reverse_json = function
   | x -> x
 
 let reverse in_filename out_filename =
-  let json = reverse_json (Yojson.Safe.from_string (contents_of_file in_filename)) in
-  let fh = open_out out_filename in
-    output_string fh (Yojson.Safe.pretty_to_string json);
+  let json = reverse_json (Yojson.Safe.from_file in_filename) in
+  let fh = open_out_bin out_filename in
+    Yojson.Safe.pretty_to_channel fh json;
     close_out fh
 
+(* You can use the file introduction.json as an example input for the "sum" and
+   "reverse" examples. *)
 let () =
   match Sys.argv with
   | [|_; "print_parse_tree"|] -> print_parse_tree ()
   | [|_; "prettyprint"|] -> prettyprint ()
   | [|_; "sum"; filename|] -> sum filename
   | [|_; "reverse"; in_filename; out_filename|] -> reverse in_filename out_filename
-  | _ -> Printf.eprintf "yojson_example: unknown command line\n"
+  | _ ->
+      Printf.eprintf
+        "%s: unknown command line\n"
+        (if Array.length Sys.argv > 0 then Sys.argv.(0) else "")
--- a/examples/introduction.ml
+++ b/examples/introduction.ml
@@ -7,10 +7,11 @@ let json_string =
 (* Print as parse tree *)
 let print_parse_tree () =
   Format.printf "%a\n" Yojson.Safe.pp (Yojson.Safe.from_string json_string)
-  
+
 (* Print as JSON *)
 let prettyprint () =
-  Printf.printf "%s\n" (Yojson.Safe.pretty_to_string (Yojson.Safe.from_string json_string))
+  Printf.printf "%s\n"
+    (Yojson.Safe.pretty_to_string (Yojson.Safe.from_string json_string))
 
 (* Read some JSON from a file, add up all numbers, and print. *)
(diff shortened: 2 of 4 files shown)


examples/introduction.ml
```diff
@@ -0,0 +1,57 @@
+(* Get a file as a string *)
+let contents_of_file filename =
+  let ch = open_in_bin filename in
+    try
```
Leonidas-from-XIV: Wouldn't this be nicer using `Fun.protect`?

examples/introduction.ml
```diff
@@ -0,0 +1,57 @@
...
+let rec sum_json : Yojson.Safe.t -> float = function
+  | `Float f -> f
+  | `Int i -> float_of_int i
+  | `String _ -> 0.
+  | `List l -> sum_floats (List.map sum_json l)
+  | `Assoc l -> sum_floats (List.map sum_json (List.map snd l))
+  | _ -> failwith "unexpected construct in sum_json"
+  
+let sum filename =
+  Printf.printf "%f\n" (sum_json (Yojson.Safe.from_string (contents_of_file filename)))
```
Leonidas-from-XIV: I think it would be nicer to use `Yojson.Safe.from_file` here. That basically does all the `contents_of_file` already.

examples/introduction.ml
```diff
@@ -0,0 +1,57 @@
...
+
+(* Read some JSON from a file, reverse any lists in it, and write to file. *)
+let rec reverse_json = function
+  | `List l -> `List (List.rev (List.map reverse_json l))
+  | `Assoc l -> `Assoc (List.map (fun (k, v) -> (k, reverse_json v)) l)
+  | x -> x
+
+let reverse in_filename out_filename =
+  let json = reverse_json (Yojson.Safe.from_string (contents_of_file in_filename)) in
```
Leonidas-from-XIV: Likewise here. One could also use `Yojson.Safe.from_channel` to show a different way of consuming input.

examples/introduction.ml
```diff
@@ -0,0 +1,57 @@
...
+  let fh = open_out out_filename in
+    output_string fh (Yojson.Safe.pretty_to_string json);
```
Leonidas-from-XIV: I think it would be nicer to use `Yojson.Safe.pretty_to_channel` and output to the channel directly, without going through the string first.

examples/introduction.ml
```diff
@@ -0,0 +1,57 @@
...
+
+(* Read some JSON from a file, reverse any lists in it, and write to file. *)
+let rec reverse_json = function
+  | `List l -> `List (List.rev (List.map reverse_json l))
+  | `Assoc l -> `Assoc (List.map (fun (k, v) -> (k, reverse_json v)) l)
+  | x -> x
+
+let reverse in_filename out_filename =
+  let json = reverse_json (Yojson.Safe.from_string (contents_of_file in_filename)) in
+  let fh = open_out out_filename in
```
Leonidas-from-XIV: Better use `open_out_bin`, to avoid end of file conversion shenanigans.

examples/introduction.ml
```diff
@@ -0,0 +1,57 @@
...
+    output_string fh (Yojson.Safe.pretty_to_string json);
+    close_out fh
+
+let () =
+  match Sys.argv with
+  | [|_; "print_parse_tree"|] -> print_parse_tree ()
+  | [|_; "prettyprint"|] -> prettyprint ()
+  | [|_; "sum"; filename|] -> sum filename
+  | [|_; "reverse"; in_filename; out_filename|] -> reverse in_filename out_filename
+  | _ -> Printf.eprintf "yojson_example: unknown command line\n"
```
Leonidas-from-XIV: In here one could match on the first element (`Sys.argv.[0]`) to read the name of the executable, saves hardcoding `yojson_example`.

examples/introduction.json
```diff
@@ -0,0 +1,3 @@
+{"Number" : 1,
+ "String" : "two",
+ "List": ["containing", "different", "types", 35.4]}
```
Leonidas-from-XIV, line 3: What is this file for? It is not used in the code as far as I see.

Leonidas-from-XIV (changes requested):
Thanks for the examples, that's nice and I'll gladly include them once ready. I appreciate the way that they are in line with the rest of the code as I would not want to install public executables onto users systems.

What would be nice is also examples in the documentation using odoc. That way users would see the examples on [ocaml.org](https://ocaml.org/p/yojson/latest) in the documentation, which is probably more approachable than the source repository.

examples/introduction.json
johnwhitington: It's to give an example input file for the "sum" and "reverse" tests, which take an input file on the command line. I have added a note to that effect in the source.

johnwhitington:
Thanks. I've fixed the issues.

You idea about examples in the documentation sounds good too for the future.

Leonidas-from-XIV:
Cool. Can you fix the remaining CI failures? It seems like it's only missing reformatting.

Leonidas-from-XIV:
Can you fix the remaining formatting failures? I tried pushing the fixes to your branch directly, but don't have permission to do so.

johnwhitington:
The commit I pushed a couple of hours ago https://github.com/ocaml-community/yojson/pull/197/commits/4165acabae9f144682deeb451a0114b5650a04a1 is still waiting for CI to report. I'm hoping the fact I used a more modern version of ocamlformat than your `.ocamlformat` suggests is ok. But we'll see when the CI reports...

Leonidas-from-XIV:
The formatting linter [has already finished](https://ocaml.ci.dev/github/ocaml-community/yojson/commit/4165acabae9f144682deeb451a0114b5650a04a1/variant/%28lint-fmt%29) and while the ocamlformat version hasn't been a problem, the `dune` file is misformatted, hence my failed attempt to push a fix directly instead of pestering contributors with minutia ;-)

The easiest way to get everything formatted correctly is just calling `dune fmt`, which should be reformatting everything that should be reformatted.

Leonidas-from-XIV:
Merged. Thanks for your contribution!

